PR #21, opened 2016-09-07
Update to include the installation via Powershell

Flowhow:
Add the information on how to turn on Shell Launcher using PowerShell (incl. Prerequs)


brianlic-msft:
We talked to the Shell Launcher engineering team and they feel that the currently documented methods are sufficient.

PR #25, opened 2016-09-13
Update provision-pcs-with-apps-and-certificates.md

eross-msft:


eross-msft:
The topic has been updated with your suggestions and should go live tomorrow morning with our daily publish around 10am.

PR #28, opened 2016-09-15
Update distribute-apps-from-your-private-store.md

imfgmbh:


brianlic-msft:
@imfgmbh Can you submit this as an issue instead?

imfgmbh:
I'll try, still fuddled by gitHub (newbie). :-)

PR #42, opened 2016-09-27
Clarify Windows versions this article applies to

Borkason:


brianlic-msft:
@Borkason Thanks for submitting this pull request!  Instead of adding specifics about editions in the applies to section, which will be challenging to keep accurate over time, we've added a link to a page that breaks down features by edition.

PR #71, opened 2016-11-09
Update create-wip-policy-using-intune.md

AndreasStenhall:
There is a bug when creating WIP policies using for instance the Swedish language Intune management portal and the policies simply do not apply in that scenario. So to prevent users from creating WIP policies and wondering why they simply do not work (I struggled hours with this before finding out the language bug) you should add this text to this article.


eross-msft:
Closing the pull request since you're working directly with the product team to figure out if this is a bug. If this ends up being more than a bug, please open another pull request and we'll get the content updated

PR #73, opened 2016-11-09
Update bitlocker-countermeasures.md

AndreasStenhall:
Good information that should be exposed to avoid confusion.


brianlic-msft:
@AndreasStenhall Sorry for the delay on this.  We're almost done rewriting this topic so I need to close this pull request for now.  If it's still missing information about the update, please create a new pull request.

PR #80, opened 2016-11-28
Update waas-branchcache.md

bryandam:
The first paragraph of the Configure clients for BranchCache section was duplicated.  I kept the version with links.  Thanks for your work!


brianlic-msft:
@bryandam Thanks for catching this!

PR #83, opened 2016-12-01
Update implement-microsoft-passport-in-your-organization.md

tonyartisan:
I am managing a server with Windows Server 2012 R2 Essentials (build 9600) and this text misled me to believe that **Turn on PIN sign-in** group policy did not affect the convenience PIN for computers running version 1607. Server-side group policies relevant to the Windows Hello for Business component are not available until Windows Server 2016. After further research and experimentation, I found that enabling **Turn on PIN sign-on** via group policy object set by the server does affect the **Turn on convenience PIN sign-in** local policy of a client running version 1607.


jdeckerMS:
Thanks for the feedback, Tony. The product team has confirmed that your change is correct. We're going to rewrite that sentence to make it even clearer.

PR #85, opened 2016-12-07
Update use-system-center-configuration-manager-to-manage-devices-with…

PaulHarfield:
…-semm.md

Running PowerShell via Configuration Manager is supported, what the script does is largely irrelevant. The statement is misleading as use of the Surface Configurator Tool in MSI format is dependant on the type of device (i.e. Surface Pro 4 or Surface book), the scripts just execute the same code in a slightly more enterprise friendly way.


eross-msft:
Thank you for your help to make this topic better, we appreciate you taking the time to contribute! At this time, we're unable to accept your change because your update isn't accurate. The current text states what Microsoft officially supports.

PR #94, opened 2016-12-23
Hint at difference in computer and user settings.

Martin-Stangl:
The user configuration contains only a subset of Windows Hello settings. 
Might be worth pointing out. Would have helped me save some time and therefore might be helpful to others as well.


eross-msft:
Thank you for your help! Your change has been accepted and the topic will be updated later today.

PR #95, opened 2017-01-03
Add line break in related topics

KeenRivals:
Separate links on lines in Related Topics to make them consistent.


eross-msft:
Thank you for your help! Your change has been accepted and the topic will be updated later today.

PR #103, opened 2017-01-09
Corrected Energy Star typo: changed "50" to "S0"

greiginsydney:


eross-msft:
Thank you for your assistance. We've accepted your change and you should see it on the site later today.

PR #108, opened 2017-01-21
Update configure-windows-10-taskbar.md

loosus456:
Just a few minor clarifications that I think may be helpful.

### DIFF
--- a/windows/manage/configure-windows-10-taskbar.md
+++ b/windows/manage/configure-windows-10-taskbar.md
@@ -10,21 +10,21 @@ localizationpriority: high
 ---
 # Configure Windows 10 taskbar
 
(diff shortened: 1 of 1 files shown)


jdeckerMS:
Thanks for your feedback! Some of your changes have been merged into the topic. A few (such as hyphenating the descriptions of the XML file and removing "by" before "using") were not because they don't comply with Microsoft content style guidelines.

loosus456:
Thanks.

Are you guys going by a mainstream style guide or an internal one?

jdeckerMS:
We have internal guidelines. For example, "by using" is preferred to simply "using" based on input from our localization experts -- it makes it easier for them to localize into other languages.

loosus456:
Is it publicly published anywhere?  It might be easier going forward with those sort of things in mind.

jdeckerMS:
We welcome input on incorrect, confusing, or missing information, but we really don't expect customers to invest their time and effort just on minor editorial issues.

PR #112, opened 2017-01-25
Update - Further Notes

PLogFinder:
Details of implementation that may be relevant to security administrators operating AppLocker to secure FOSS applications certified by Certum certificate holders.

These "further notes" really belong as a comment under the article, but I ended up here, so contributing, for your consideration.


Justinha:
@PLogFinder Thanks for your submission and sorry for the delayed reply. Our engineering team indicated that the statement about what Applocker is doing is not accurate, so we have not added it to the topic. Somewhat related, our new Docs.microsoft.com publishing platform allows comments at the end of articles.

PR #125, opened 2017-02-16
Update domain-controller-ldap-server-signing-requirements.md

tvuylsteke:
This article, and also the "legacy" one (https://technet.microsoft.com/en-us/library/jj852234(v=ws.11).aspx ) suffer both the confusing wording covering the relation between this setting and LDAP Simple Binds. I blogged about this a while ago: http://setspn.blogspot.be/2016/09/domain-controller-ldap-server-signing.html

And now I'm trying to get this one corrected. Made an edit. But maybe there's still room for improvement.


Justinha:
Unfortunately we don't have a way to revise the "legacy" version.

PR #140, opened 2017-03-14

FrancescoPapini77:
https://blogs.technet.microsoft.com/mniehaus/2016/08/16/windows-10-delivery-optimization-and-wsus-take-2/
The “LAN” mode identifies PCs that are on the same LAN by looking at their external IP address – all PCs going through the same internet IP (through a proxy server or router) are considered to be on the same “LAN.”


DaniHalfin:
Hi @FrancescoPapini77. Thanks for the input. Looking into how to best phrase it. You should see a change to this topic soon.

DaniHalfin:
Quick update @FrancescoPapini77 - we have some additional content coming out soon (no exact ETA yet) for Delivery Optimization. We would like to wait with this change so that we can reference this content to give more details on how this works.

PR #145, opened 2017-03-22
Corrects default notification values

exchange12rocks:
In modern OSes they have been changed to 5 days


Justinha:
@exchange12rocks Thanks for the correction!

PR #162, opened 2017-04-14
Added caveat for line 110: group membership in offline migration

jmullman:
Per [Offline Migration Reference](https://technet.microsoft.com/en-us/itpro/windows/deploy/offline-migration-reference), user-group memberships are *not* migrated by default. Use of the **&lt;ProfileControl&gt;** section in Config.xml is required.


greg-lindsay:
Thanks for the note I will update the docs

PR #169, opened 2017-04-23
Removed extra spaces for accuracy/consistency

npherson:
When referencing KB article numbers, there isn't a space, e.g. KB12345 instead of KB 12345.


eross-msft:
@npherson 
We've recently moved to a new content location, going from TechNet to docs.microsoft.com. In the process, we significantly updated our content structure, which has caused some conflicts with your fix.
To get around these conflicts, I need to close this pull request and have you open a new one with the same fix.
As soon as you make the change, I'll be able to merge it into our content and have it go live.
Thanks!!

PR #171, opened 2017-04-25
Updated bad link for Group Policy templates

npherson:
Now links to the latest Group Policy Download...


jamiejdt:
Thanks for catching and correcting this. LGTM

PR #179, opened 2017-05-02
Unintended usmt migration

kaushika-msft:
The usmt from the adk for windows 10 is meant to migrate settings from these lower operating systems to windows 10 and not among each other.


lizap:
@kaushika-msft - thanks, Kaushik! We'll get this merged today. - Liza

PR #195, opened 2017-05-12
Update windows-spotlight.md

loosus456:
Added 1703 warning for lock-screen Group Policy.


jdeckerms:
Thanks for your suggestion. I need to confirm the change with the product group before accepting it.

PR #210, opened 2017-06-19
Windows Store and LTSB version clarity

kaushika-msft:
Making clear statement as Windows Store apps is not supported on Windows 10 LTSB version.


lizap:
#sign-off Ok to merge.

PR #211, opened 2017-06-20
Updated to include fsiAnyCpu.exe

enigma0x3:
Same as FSI.exe, has different fileName.


Justinha:
@enigma0x3 Thanks for the contribution. I'm confirming the change with our engineering team.

PR #219, opened 2017-06-26
Update column width on SupportedProductPlatform

strican:
Fixing a column width that wasn't aligned with the rest of the page.


MariciaAlforque:
Ready to merge.

PR #220, opened 2017-06-28
Ntkd kernel debugger addition

enigma0x3:
ntkd is a kernel debugger that is nearly identical to the already blocked kd.exe binary.


Justinha:
@enigma0x3 Thanks for your contributions. We are having them reviewed by the engineering team.

PR #221, opened 2017-06-28
Update enable-cloud-protection-windows-defender-antivirus.md

arottem:
Note to editor: While this document covers all version of Windows 10 (hence the update to the doc), recommend creating a similar for Win7 (or at least note that it is possible)


iaanw:
Merge approved

PR #234, opened 2017-07-17
TPM 2.0 is now the main recommendation

computeronix:
TPM 2.0 (especially for enterprises - and starting with 1607 the number of additional capabilities that requires this)

TPM 2.0 will help prevent (reduce) lockout attempts and provide more efficiencies with Windows Hello for Business.  THis should be the main requirement now moving forward especially with 1607 and later.


DaniHalfin:
Hey, it'll take us a few days to talk about this. I'll respond here ASAP.

DaniHalfin:
Hi again, upon further discussions we've decided to keep it as it is. While TPM 2.0 is recommended , this is still official documentation and we have document all supported TPM version.

PR #237, opened 2017-07-17
Update tpm-recommendations.md

computeronix:
Modified Device Guard line to match https://docs.microsoft.com/en-us/windows/device-security/device-guard/requirements-and-deployment-planning-guidelines-for-device-guard


Justinha:
@computeronix thanks for your contribution.

PR #238, opened 2017-07-17
Update stop-employees-from-using-the-windows-store.md

computeronix:
Windows 10 1511 is no longer needed to be specified.  As of 1511 Pro is no longer supported but Enterprise and Education continue to be supported

This will better articulate to enterprise customers sayings its there and can be used (on all versions of windows 10) while specifically mentioning the Pro edition does not support this starting with the specific build.

Updated the blurb to specify this as well and link to the KB for additional information (as mentioned: https://docs.microsoft.com/en-us/windows/client-management/group-policies-for-enterprise-and-education-editions)


TrudyHa:
Thanks for your contribution. We'll update the topic.

PR #242, opened 2017-07-18
Bad links and Upgrade Analytics Adjustment

computeronix:
The Upgrade Analytics link has been updated and fixed to point to Upgrade Readiness
Also renamed Analytics to Upgrade Readiness (which is the new name for the capability being discussed in this section)

Corrected Block at First Sight link


jaimeo:
I see no reason this could not be merged, though I would like to view the preview of it first....

PR #248, opened 2017-07-24
Update microsoft-surface-data-eraser.md

BrutalCoding:
Just a very small typo fixed regarding the Surface Pro 3


BrutalCoding:
Also there's another thing, it is about the name "Surface Pro":

> [!NOTE] Third-party devices, Surface devices running Windows RT (including Surface and Surface 2), and Surface Pro are not compatible with Microsoft Surface Data Eraser.

It might confuse people to just call it "Surface Pro". I assume that you referred to the Surface Pro from 2013, but the new Surface Pro from this year, 2017, has the same name.

PR #257, opened 2017-08-07
Fix broken tip

shuurajou:
Hopefully this fixes the broken tip.


jdeckerms:
Thanks for noticing the issue and making the fix!

PR #261, opened 2017-08-11
Remove git metadata from vda-subscription-activation

gsacavdm:


greg-lindsay:
Seems there was a merge conflict apparently. I think the changes described will fix it but I need to verify.

PR #270, opened 2017-08-23
As per CI bug 66158

andreiztm:
"Long-Term Servicing" instead of "Semi-Annual" in the last table for the "Switching channels" section


DaniHalfin:
Hi Andrei, good catch. I'm not sure how I missed this. 
There's actually more fixing that needs to happen here, so I'm going to not merge this one, but I will fix rewrite the parts you highlighted so that they fully make sense (and don't say Semi-Annual Channel for Business, for example).

Thanks!

andreiztm:
Hi Dani,

Thanks, yes I wasn't too sure about that one myself :P

Have a nice day,
Andrei

DaniHalfin:
Fixed! :) Thanks again.

PR #273, opened 2017-08-26
Consolidation on the "Summary" section

skycommand:
(1) The "Disk validation" section is moved up, right below the "Summary" section and is called "Prerequisites". Reason: It is so important that must not be buried. (2) Consolidated all elements of what MBR2GPT can do under a pre-existing bullet list and removed repetitions. (3) Created a list of what MBR2GPT cannot do.

### DIFF
--- a/windows/deployment/mbr-to-gpt.md
+++ b/windows/deployment/mbr-to-gpt.md
@@ -17,29 +17,41 @@ ms.localizationpriority: high
 
 ## Summary
 
-**MBR2GPT.EXE** converts a disk from Master Boot Record (MBR) to GUID Partition Table (GPT) partition style without modifying or deleting data on the disk. The tool is designed to be run from a Windows Preinstallation Environment (Windows PE) command prompt, but can also be run from the full Windows 10 operating system (OS).
+**MBR2GPT.EXE** converts a disk from the Master Boot Record (MBR) partition style to the GUID Partition Table (GPT) partition style without modifying or deleting data on the disk. The tool is designed to be run from a Windows Preinstallation Environment (Windows PE), but can also be forced to run from the full Windows 10 operating system (OS).
 
-MBR2GPT.EXE is located in the **Windows\\System32** directory on a Windows 10 computer running Windows 10 version 1703 or later.
+MBR2GPT.EXE is located in the **Windows\\System32** folder on an instance of Windows 10 version 1703 (also known as Creators Update) or later.
 
-You can use MBR2GPT to perform the following:
+MBR2GPT can perform the following:
 
-- \[Within the Windows PE environment\]: Convert any attached MBR-formatted system disk to the GPT partition format.
-- \[From within the currently running OS\]: Convert any attached MBR-formatted system disk to the GPT partition format.
+- Conversion of an attached disk containing a copy of Windows 10 version 1703 from the MBR partition style to GPT.
+- Offline conversion of an attached disk containing a copy of Windows 10 versions 1507, 1511 or 1607 from the MBR partition style to GPT. 
+- Offline conversion of an attached disk containing a copy of Windows 7, 8 or 8.1 from the MBR partition style to GPT. However, such a conversion is not officially supported and its reliability is not guaranteed. The recommended method to convert these disks is to upgrade the operating system to Windows 10 first, then perform the MBR to GPT conversion.
 
->MBR2GPT is available in Windows 10 version 1703, also known as Windows 10 Creator's Update, and later versions. 
->The tool is available in both the full OS environment and Windows PE. 
+MBR2GPT **cannot** perform the following:
+
+- Convert an arbitrary disk without a Windows operating system.
 
 You can use MBR2GPT to convert an MBR disk with BitLocker-encrypted volumes as long as protection has been suspended. To resume BitLocker after conversion, you will need to delete the existing protectors and recreate them.
-
-The MBR2GPT tool can convert operating system disks that have earlier versions of Windows 10 installed, such as versions 1507, 1511, and 1607. However, you must run the tool while booted into Windows 10 version 1703 or later, and perform an offline conversion.
-
-Offline conversion of system disks with earlier versions of Windows installed, such as Windows 7, 8, or 8.1 are not officially supported. The recommended method to convert these disks is to upgrade the operating system to Windows 10 first, then perform the MBR to GPT conversion.
 
 >[!IMPORTANT]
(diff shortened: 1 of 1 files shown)


DaniHalfin:
Hi @skycommand. 
The location\consolidation changes will be reviewed in a while. Knowing that a list of what MBR2GPT can't do is needed is great feedback. We'll review that as well.
As for the stylistic changes - that's the personal preference of the writer, and I don't see a reason to prefer one way over that other. In some case, these changes are inconsistent with the language used in the rest of the library. With that in mind, stylistic changes won't be implemented.
Thank you for your contribution.

greg-lindsay:
Some of the stylistic changes are an improvement, others are not. We shouldn't call things folders when referring to a command prompt. They are directories.  When we say the app can be forced to run under the full version of Windows I think this is because it's necessary to add the "-force" option.

skycommand:
@greg-lindsay  Hi. Your point about "folder" is probably fair. But, skipping the prerequisite change and the list of what MBR2GPT cannot do, AFAIK that's the only style change I made. (Your point about "forced" is actually my contribution, not my deviation.) So we are left with:

- Removal of "command prompt": That was factually inaccurate. One can run MBR2GPT from PowerShell or even from Run... dialog box. Since Windows 10 is bringing PowerShell to the front-line, this "command prompt" restricting elements is definitely taken to mean "but not from PowerShell!"
- Addition of "the": It is a grammar point. Name of computer software, web services, software components and software technologies (which are all products of a source code) need no definite article **only if they appear alone**, e.g. "I run Windows". But a descriptor makes "the" necessary, e.g. "I run the Windows operating system." Hence, it is either "GUID Partition Table" or "the GUID Partition Table partition style".

There is of course the sentence "You can use MBR2GPT to perform the following" which I turned into "MBR2GPT can perform the following". If you don't consider it part of the consolidation efforts, yes, it can be seen as both another style change or a semantic change. "You can" is a meaning-restricting element that reduces accuracy: Not just you, but anyone or any non-human entity can do this with MBR2GPT. Still, since the reduction in accuracy is negligible, it leaps to the eye that avoiding second person pronoun in technical articles is a well-established matter of style.

greg-lindsay:
Hi - I am reviewing the changes again now that I'm in the office.  
I like the idea of moving the disk validation section up and renaming it to prerequisites, and I've done this. 
Specifying "the" Master Boot Record partition style instead of just saying Master Boot Record is a solid correction. 
I thought I had written "forced to" run but I see this is a requested addition. I was going to change that phrasing if I had written it, so I'll do that now by adding something more specific to the end of the sentence.
The next set of changes aren't really needed for clarity but I see that Windows 10 was used twice here so one instance should be eliminated (I had written a Windows 10 computer running Windows 10, which is a silly typo due to rewriting that sentence improperly).
The edit "MBR2GPT can perform the following" is too anthropomorphic.
The next change is good because it points out that the two bullets are essentially saying the same thing. Originally the two statements were different but an edit made them the same. I should have gone back and rewritten this so it wasn't redundant. However, the proposed changes are not correct because data disks are not supported. This gives me an opportunity to rewrite this section and point out that fact. Non-system disks might be supported in the future. We should not refer to disks as "arbitrary."
Thanks for all the suggestions on this page.  I am pushing out a rewrite today.

greg-lindsay:
Also, in reference to running MBR2GPT from a PowerShell prompt. It's probably possible to do it. You would likely need to escape some of the command line parameters. It isn't the proper way to run a command line tool. The proper way to run a command line tool from PowerShell is to preface it with cmd /c which essentially runs it from a command prompt within PowerShell.

skycommand:
@greg-lindsay Sounds like you've approved most of my changes by now. I can only be happy.

On a second thought I agree: The word "arbitrary" is wordiness. I could simply write "Convert a disk without a Windows operating system." Of course, on the same vein, "data disk", which you suggested, would be wordiness too.

You say "perform" is anthropomorphic? Well, try search it in the Corpus of Contemporary American English (COCA). Sorry for not posting a link; COCA is a web app and is not link-friendly. Anyway, the word "performance" is vastly used by Microsoft. Why not "perform"?

greg-lindsay:
@skycommand I want to just say again that I really appreciate all the suggestions you had. The changes have improved the article and fixed some silly errors, so thanks a ton!  

W.r.t. perform it is just something my editor hammered into me when I started as a writer. Other writers might not have gotten the same advice, so it's possible you could see an application performing actions in some other articles.  But for a moment let's assume that a picture was hung in the hallway. Did a hammer hang the picture, or a person? This is obviously stretching the concept, but you get the idea. The article is extolling the virtues of MBR2GPT so it might be OK to talk about all the actions it can perform without sounding odd, but I like to place the power per se on the person using the tool rather than the tool itself. It's a minor point for sure.

skycommand:
@greg-lindsay Oh, yes. I remember that certain trend you are talking about! (With a pang of nostalgia.) It was wrong to say "armies will shatter, worlds will burn!" because neither do armies shatter for no reason nor do worlds burst into flames spontaneously. It was correct to write "armies will be shattered, worlds will be burned" to imply that we are deliberately withholding the subject of these heinous misdeeds. (Alternatively, include a subject and write in active voice.)

But this is no longer the tradition these days. The counterargument put forth today is that "you don't see 'spontaneously' in the original sentence, hence your assumption about it being there is your own. The sentence simply says 'armies will shatter' and does not explain why." I sincerely suspect this has come from Wikipedia. You see, there is a lot of pressure on editors there to neither assert nor imply something outside facts and not to use weasel words either. So, they write in active voice using intransitive verbs just to report the fact. "In 1998, two of the jungle areas burned." We don't know why and it does not matter, so the writer didn't imply or assert it as a heinous misdeed or otherwise.

As for the "Did a hammer hang the picture, or a person?", please be careful: there is a metonymy hidden in there! It can induce wrong conclusions. A hammer has a tiny but important role in a whole procedure leading to a picture being hung on a wall. MBR2GPT pretty much does everything; the only role a human has in it is giving the order.

greg-lindsay:
You bring up some interesting points. This reminds me of an editorial change where I pushed back heartily. Consider these two similar sentences:

You can convert a system disk from MBR to GPT by using MBR2GPT.
You can convert a system disk from MBR to GPT using MBR2GPT.

Of course there are other options such as using "with" but let's ignore that for now. Which of the two do you prefer and think is correct? I preferred the second one as a general rule, but my editor had added "by" to every instance of "using" in an article I wrote several years ago. 

The difference comes down to where you assign the power. To assign power to the tool, you say "using X you can achieve the impossible." To assign power to the user, you say "by using X you can achieve the impossible."

Usually I just go with what sounds better and don't think about it too much.
In the current doc I ended up removing the word "perform" and just saying you can use MBR2GPT to: (list).

I appreciate the discussion.

PR #274, opened 2017-08-26
Fix incorrect directory

weijiechai:
Directory will be created in "C:\Users\administrator.CONTOSO" instead of "C:\" without backslashes.


DaniHalfin:
Looks good. Thanks for the catch @weijiechai !

PR #275, opened 2017-08-31
Grammar fix and factual accuracy

skycommand:
1. **Grammar fix:** Names of computer programs do not require the definite article ("the") when they appear alone. (e.g. "I like Windows 10", not "I like the Windows 10". But the definite article appears when the name become part of a noun group, e.g. "I like the Windows 10 operating system".
2. **Corrected factual inaccuracies:** For example WDSC neither checks for third-party firewalls nor relies on the "Windows Security Center" service; it uses its own "Windows Defender Security Center Service".


iaanw:
Thanks for these suggestions and feedback! We're currently working on an update to this topic/library as part of the Fall Creator's update and we'll include your feedback in those changes.

PR #283, opened 2017-09-12
Fix-Manage-Connections-Registry-Key

cryps1s:
Validated that the PreventDeviceMetadataFromNetwork value in the documentation is incorrect on a fresh install of 1703. Updated documentation to reflect the correct location.


Justinha:
@cryps1s Thanks for your submission. We validated the published registry path with our test team. When Group Policy is set it will create the corresponding registry entry. On a machine running Windows 10, version 1703 without the GP configuration, the entry appears at the path you specified.

cryps1s:
Aha, good catch. Just validated in my lab -- I'll close out this PR. Thanks!

PR #290, opened 2017-09-20
Update upgrade-readiness-get-started.md

nevedita:
Fixing the End Point which need to be whitelisted for Vortex Connectivity and Settings.


wayneroberts386:
Unsubscribe

PR #294, opened 2017-09-21
Fixing the hyperlink for 0xC1900101 article

kaushika-msft:
Hyperlink broken due to typo.
[0xC1900101 error when Windows 10 upgrade fails after the second system restart'(https://support.microsoft.com/help/3208485/0xc1900101-error-when-windows-10-upgrade-fails-after-the-second-system)


lizap:
Thanks, Kaushik!

PR #300, opened 2017-10-04

mtniehaus:
There was a reference to Windows 10, version 1611 in the doc - there is no such thing.  I assume it should have been 1511, made that change.  If it needs to be something else (e.g. 1607) then feel free to correct.


nickbrower:
Thanks for your contribution Michael!  I tracked down the correct version today: 1607.  If you would like to update your pull request with the correct version, we will gladly accept the change.

mtniehaus:
OK, I’ve updated it to 1607 using the same pull request.

Thanks,
-Michael

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10
[HxS - 16299 - 16.0.8700.4009]

PR #305, opened 2017-10-12
Update upgrade-readiness-additional-insights.md

zdvorak:
Minor fixes to clarify that Site discovery does work on Win10 machines now.


jaimeo:
This can go ahead and go live as far as I'm concerned.

PR #314, opened 2017-10-19
Line 32 referes to a boken url

y0avb:


jdeckerms:
Thanks, Yoav, we'll merge this and push it live.

PR #316, opened 2017-10-23
Update get-support-for-security-baselines.md

karimelmel:
Corrected SCT 4.0 to SCM 4.0 in Security Tools for Microsoft Products.


Justinha:
@karimelmel Thanks for the correction!

PR #320, opened 2017-10-25
Fix invalid name

sylveon:
image


iaanw:
Nice catch, thanks! Will get this merged in.

PR #321, opened 2017-10-26
Update upgrade-readiness-additional-insights.md

zdvorak:
Added a disclaimer to say that Edge data will be collected starting with the newest Insider preview builds.


jaimeo:
Looks great; please proceed with merge.

PR #335, opened 2017-11-08
Update upgrade-readiness-requirements.md

nevedita:
Adding one more endpoint which needs to be whitelisted


jaimeo:
Looks great; please merge.

PR #338, opened 2017-11-12
Update response-actions-windows-defender-advanced-threat-protection.md

WRCko:
Sentence "These response actions are only available for machines on Windows 10, version  1703" is no longer valid since these functions and many more can be used with Windows 10 1709. I suggest to add "or higher" into sentence. Thanks!


mjcaparas:
Yes, correct. This pull can be integrated.

PR #339, opened 2017-11-13
Update remotewipe-csp.md

joscon:


nickbrower:
To be consistent with other node descriptions: "Added in Windows 10 Fall Creators Edition" -> "Added in Windows 10, version 1709".  Otherwise perfect, so thank you!

PR #343, opened 2017-11-16
Update index.md

jmunck:
Added Tip to note that skipping a version (1511 - 1709) is indeed possible.


jaimeo:
Your addition of the tip is fine (though double-check that it renders properly in Staging), however, you deleted the paragraph with information and link about Feedback Hub. Please replace that. Also your sentence needs to end in a period. And preceding that, there should only be ONE space after all punctuation, including periods.

jaimeo:
Thanks for fixing those details. Persons with the appropriate perms, please go ahead and merge.

PR #348, opened 2017-11-20
Update add-profile-to-devices.md

computeronix:
Modified the line about the account rep and provided the location of the powershell script to allow users to self run this and generate the CSV

Please turn the URL into a microsoft go link so as this link changes it can easily be updated.


computeronix:
modified wording

PR #359, opened 2017-12-01
Update hello-cert-trust-adfs.md

BroadArc:
certutil.exe command does not work on Windows Server 2016. It worked on my Windows 10 1709 build.


MikeStephens-MS:
Thank you for the request.  When following the deployment guide, the command should be ran from the newly updated ADFS 2016 server.   The introductory sentence under the "Mark the template as the Windows Hello sign-in template" states "Sign-in to an **AD FS Windows Server 2016** computer with Enterprise Admin equivalent credentials."  If you receive this error, its because the AD FS server does not have the required updates, or has not been rebooted after the updates were applied.

PR #365, opened 2017-12-07
Remove notes about using browsers that support HPKP

AngeloKai:
Chrome is deprecating HPKP: https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/he9tr7p3rZ8. Edge and Safari never supported it because the feature had low adoption.


MikeStephens-MS:
We will remove reference to HPKP, rather than the suggestion included here.  There's not point in telling people something is going away and not telling them what is replacing it and when-- it just creates more questions that do not have answers.  We will leave it at Enterprise pinning is not meant for public names and leave it at that

eross-msft:
Closing pull request

PR #373, opened 2017-12-11
Update upgrade-readiness-get-started.md

nevedita:
Changing the end points that need to be whitelisted to the correct ones.


nevedita:
Some typos got introduced. Will open a new pull request.

PR #379, opened 2017-12-14
Update hello-biometrics-in-enterprise.md

mapalko:
Anti-spoofing for fingerprint is currently not configurable and is a requirement for all sensors. This document is being updated to reflect that and to remove references to FRR without anti-spoofing for fingerprint.


brianlic-msft:
@mapalko Let me know when you were able to make those changes and I'll merge the PR.

mapalko:
@MikeStephens-MS employee is used extensively throughout this document. I feel like that should be a separate change since this one is very scoped.

PR #384, opened 2017-12-16
Finding the right Registrar Pool

y0avb:
If running the existing command, you'll get: 
PS C:\WINDOWS\system32> (Get-CsTenant).TenantPoolExtension
MCOProfessional:sippoolDB42E08.infra.lync.com

If running the right command, you'll get:
PS C:\WINDOWS\system32> (Get-CsTenant).RegistrarPool
sippoolDB42E08.infra.lync.com

Otherwise, the script will forever fail.


jdatms:
Thanks, Yoav!

PR #396, opened 2017-12-22
Update start-layout-xml-desktop

alhopper-msft:
The ability for OEMs to pin apps to the MFU section of Start was deprecated in 1709. OEMs can still include the TopMFUApps tags in the layoutmodification.xml, but it will have no impact to the Start layout.


DaniHalfin:
@jdeckerms @eross-msft

jdatms:
Thanks for letting me know about this change, Allison! I'm not going to merge this change, because I don't want to delete the info, as business customers on previous versions might still need it. Instead, I'll add your note and also will add clarification in each row that those elements are only for versions prior to 1709.

eross-msft:
Closing pull request.

PR #407, opened 2018-01-05
Update customize-exploit-protection.md

markwoMSFT:
Add notes for anti-ROP mitigations indicating they are not compatible with ACG. Users that enable ACG for a process should not enable the anti-ROP mitigations at the same time.


mjcaparas:
verified accurate. ok to merge.

PR #430, opened 2018-01-18
added a tip ([!Tip] It is not possible to let the String empty between:)

slaet:


brianlic-msft:
@jdeckerms Can you take a look? Once you approve it, I'll merge it in.

jdatms:
@slaet Can you confirm that I'm interpreting this correctly? You're adding a tip on how to remove default pinned apps and not pin any new ones?

slaet:
Yes, that is correct.

jdatms:
Thanks for clarifying, @slaet. Rather than make this a tip in the section for "Remove default apps and add your own", I think it would be better as its own section (to help customers find it). I'll revise the topic on my end to add the new section.

Markdown tip, just fyi: any notes (NOTE, TIP, IMPORTANT, WARNING) must be preceded with > and each line of content also must be preceded with >    

Example:
>[!TIP]
>It is not possible...

slaet:
Great, thank you, and thanks for the TIP 😉

PR #472, opened 2018-02-06
Added branch option

lizap:


brianlic-msft:
@lizap Our guidance is to create a fork so the branch option is not available.

PR #487, opened 2018-02-12
Added Downgrade Information and Fixes to upgrade

computeronix:
This contains minor typos to the upgrade editions table (removed unneeded parenthesis).

Also added a Downgrade section discussing moving from Windows 10 Enterprise to Pro for example.

If this downgrade section makes sense and if approved by the Product group I can add the additional supported downgrade options?


greg-lindsay:
I think we added a separate doc so this isn't needed any more.

PR #506, opened 2018-02-21
Update mandatory-user-profile.md

RyanBijkerk:
The package also needs to be removed for all users otherwise the error will remain during the sysprep.


jdeckerms:
Thanks for your contribution, Ryan!

PR #592, opened 2018-03-22

marcshep-msft:
Vortex endpoint name tweak: Telemetry->Diagnostic
Clarified that KB3150513 is optional
Clarified that SSL inspection must be disabled


lizap:
Thanks, Marc!

PR #596, opened 2018-03-25

marcshep-msft:
Fixed spelling error (compatibitly->compatibility).
Clarified that windows 10 contains the compatibility update so all they need to do is keep installing cumulative updates to get the latest.
Added info on the new (1803) device name opt-in.


lizap:
Hi, Marc - can we separate out the 1803 update? We can't publish that bit until GA (although we might be able to on 4/10). Let me know if you want to discuss. Thanks, Liza

PR #600, opened 2018-03-26
Proposing we change the name of this article

npherson:
Since this is the overview of telemetry and the first page we send people to, I'm proposing we give it a better title.


brianlic-msft:
@npherson Thank you for the suggestion! Unfortunately, we can't change the name right now as major changes to this topic have to go through a legal review.

PR #601, opened 2018-03-26
clarify wip doc

cloudhandler:
clarifies selective wipe capability as specific to mdm enrolled devices only


Justinha:
@cloudhandler Thanks for the correction!

PR #613, opened 2018-03-28
Update policy-csp-power.md

ShintakuNobuhiro:
Add footnote "3" (Added in Windows 10, version 1709) to check mark.


Justinha:
@ShintakuNobuhiro Thanks for this correction!

PR #638, opened 2018-04-02
Fix typo for box 5b - Change AD FA RA to AD FS RA

nathanmcnulty:
Just fixing a typo


Justinha:
@nathanmcnulty thanks for the correction!

PR #676, opened 2018-04-09
Update index.md

heatherpoulsen:
Updated "Windows as a Service" to "Windows as a service" (small "s") per branding guidelines and updated "Windows TechCenter" to "Windows IT Pro Center" as that is what the site is called now.


lizap:
Please wait to merge while JaimeO follows up to determine scope of related changes.

PR #687, opened 2018-04-10
Update wd-app-guard-overview.md

qrscharmed:
Adding 1803 context for Pro users.


Justinha:
@qrscharmed Thanks for these additions!

PR #749, opened 2018-04-18
Update add-profile-to-devices.md

mtniehaus:
Based on feedback from HP, added text to make it clear that the PKID value is optional and typically blank.  Also shortened "Windows AutoPilot Deployment Program" references to just "Windows AutoPilot" which is more appropriate in this context.

Also, as an FYI:  The store team indicated that they are in the process of adding a new wadp014 error to indicate that the device being uploaded is already present in AAD.  That implies that there is also a wadp011, 012, and 013, so the table of errors probably needs more updates.


lizap:
Thanks, Michael!

PR #815, opened 2018-04-30
changed DISM operating system uninstall command-line options hyperlink

sandytsang:
the hyperlink for "DISM operating system uninstall command-line options" is wrong


sandytsang:
I don't understand why you closed this pull request, the link is still wrong. The review.doc.microsoft.com is for Microsoft internal use, not for public. We don't have access open that link.

PR #851, opened 2018-05-04
Update faq-wd-app-guard.md

ronasong:
Added vGPU q/a


Justinha:
@ronasong Thanks for adding this!

PR #852, opened 2018-05-04
Update configure-wd-app-guard.md

ronasong:
Updated note to vgpu


Justinha:
@ronasong Thanks again, this looks great!

PR #866, opened 2018-05-07
Proposing clarification to proxy and SSL text.

MatthewMWR:
Proposing clarification to proxy and SSL text. They were intermingled oddly before. I think addressing them separately is more logical for IT.


MatthewMWR:
First PR for docs. I hope I did this right. Jaime are you seeing this?

PR #870, opened 2018-05-07
Update Intune's configurability

arottem:
1. Update Intune's configurability
2. Remove BaFS reference as it's confusing - BaFS is available in all Windows 10 versions, and need no config. The config option is only to disable it.


Justinha:
@arottem Thanks for these corrections! They should appear later this morning.

PR #904, opened 2018-05-11
Update mbam-25-supported-configurations.md

jadye527:
It is unclear the servicing updates are required for SQL 2016 support etc....


jadye527:
Suggestions and approval for adding servicing update for MDOP as a requirement for SQL 2016 support and that the most recent servicing update should always be used.  Lastly a link to where this can be found would be great.

PR #908, opened 2018-05-12
Cleaned "-" that was breaking xml

jaromirk:
There was "-" leftover that was breaking xml syntax.


jaromirk:
Here is breaking commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/dfddbf78ef31b3741c03de0e5ae5211e04cd11a3#diff-540ffe49bd32e95e8b2a5da5ba1a565e

PR #991, opened 2018-05-30
Fix wrong HKLM keys that were missing SOFTWARE at the root

borigas:
A bunch of keys I was trying to use didn't exist. Googling the keys, it looks like they were missing `SOFTWARE\` at the root.


Justinha:
@borigas thanks for the corrections!

PR #1061, opened 2018-06-11
Update supl-csp.md

danhwang1:
We have recently made a change in our Location Platform pertaining to SUPL to increase the max number of root certificates from 3 to 6 (as mandated). As a result, we will need to update the necessary public documentation here: https://docs.microsoft.com/en-us/windows/client-management/mdm/supl-csp


lizap:
@MariciaAlforque - another CSP proposed change - can you please review? Thanks!

PR #1076, opened 2018-06-13
Corrected examples XML syntax

EagleIJoe:
Upper case boolean values caused parser error
Ending XML closing tag invalidates file


lizap:
Thanks!

PR #1090, opened 2018-06-14
Update windows-10-start-layout-options-and-policies.md

JohnRajunas:
I think adding the reference to CopyProfile not being supported will help insure IT Pros do not consider using it as a alternative to the options detailed here


jdeckerms:
Thanks, @JohnRajunas this will go live sometime today.

JohnRajunas:
Excellent – Thanks!

PR #1136, opened 2018-06-21
Correct Verb Tense in replace-a-windows-7-client-with-windows-10-using-configuration-manager.md

markekraus:


jdeckerms:
@markekraus Good catch, and thanks for taking the time to fix it!

PR #1160, opened 2018-06-26
Update whats-new-windows-10-version-1507-and-1511.md

shortpatti:


johncapital1consultants:
i'm not sure about computer stuff but call me but i''v been getting hacked so i've been pretty picky about who calls and who i give my infomation too

PR #1175, opened 2018-06-27
fix typo in install precedence listing

williamtdr:


jdeckerms:
@williamtdr Thanks for catching that!

PR #1229, opened 2018-07-03
Add UWF for IoTCore

parameshbabu:
Add UWF for IoTCore in the summary table.


jdeckerms:
@parameshbabu Thanks for catching that and making the correction!

PR #1232, opened 2018-07-03
Update manage-settings-with-mdm-for-surface-hub.md

rikot:
In the Supported Windows setting section, the link in the first paragraph for Configuration service provider reference is outdated, but does get redirected to the correct webpage.  Corrected link so that it points directly to the article


jasongroce:
Thanks @rikot ! I modified the change slightly to remove the /en-us/ from the link, so that any localized versions of the content will take the reader to their specified language if available. We really appreciate the fix!

PR #1243, opened 2018-07-05
Tweaks to reference section

marcshep-msft:
Added a link to Trust Center, removed "en-us" from link to Trusted Cloud


jaimeo:
This looks fine; please move along to master and live (I don't have the perms to do so).

PR #1262, opened 2018-07-10
Fixing default registry value set for DisableTelemetryOptInChangeNotification

Ar5ham:
This change, fixes the "Registry" section for "Diagnostic opt-in change notifications". The "GP/MDM" sections contradict the "Registry" section of the document. The default value (i.e. Enable mentioned in GP/MDM section) is 0 and not 1, setting the registry key to 1 will actually disable the notification, while the default behavior is to have it not set OR enabled (i.e. 0).


Ar5ham:
@DaniHalfin can you review this change please?

DaniHalfin:
This looks good. I've confirmed with the owner, just in case :)
Thanks for noticing @Ar5ham .

DaniHalfin:
@lizap @shortpatti ok to merge

PR #1264, opened 2018-07-10
Update policy-csp-controlpolicyconflict.md

MSFT-Kathy:
Change description field for RS5. In RS5, delete command and set the value to 0 after setting it to be 1 will be supported. Please review the text with Peter Kaufman.


MSFT-Kathy:
Suggested change is done.

PR #1304, opened 2018-07-19
Update set-home-button-url-include.md

okieselbach:
Hi,
I think there is a type or copy/paste error:
MDM Setting ./Vendor/MSFT/Policy/Config/Browser/SetHomeButtonURL has Data type: String
Integer doesn't make sense, it needs the URL to point to.
best,
Oliver


shortpatti:
@okieselbach, thank you for catching that typo.

PR #1314, opened 2018-07-20
Revert "Add an additional question to the FAQ"

Justinha:
We don't want this merged just yet. I requested changes from also want engineering to review.


Justinha:
@kenjohnson03 This is the reverted pull request. I want to hold off publishing until we can clarify that last sentence and add what is the purpose for this account, and what an IT Pro needs to know about it.

Justinha:
We need more info before we can publish this.

PR #1365, opened 2018-07-27
added MDM reference for Fallbacksources in WDAV

computeronix:
MDM is not mentioned as a possible management point for WDAV Sources.  This adds the MDM part and the policy to reference to configure the fall back sources.


johncapital1consultants:
can someone help me b/c i've been getting hacked have no clue what going on

PR #1419, opened 2018-08-02
Update reboot-csp.md

gkomatsu:
Updated the date format to match with what the CSP actually returns to prevent confusion.


MariciaAlforque:
Approved to publish.

PR #1509, opened 2018-08-21
Removing Windows Server from DH onboarding doc

MatthewMWR:
This was meant to be removed ages ago, but I just noticed it is still there.


shortpatti:
Done.

PR #1542, opened 2018-08-27
Correct Exploit Guard and Edition Requirements

computeronix:
Correcting the chart as of 1803 (HVCI in Pro) and also moving ASR to E5 only.


shortpatti:
@andreabichsel, can you review this PR?

PR #1622, opened 2018-09-12
Update docs to document new security property

andreiztm:
mbe-control

Please see the attached screenshot documenting the new option.

Thank you,
Andrei


andreiztm:
@Justinha np :) Thanks as well.

Issue #1480, opened 2018-08-16
Documentation update


tdogdfw:
I use a Surface Pro 4 with Application Guard and it works fine.  Your documentation says a hardware requirement is a 64bit processor with 4 Cores... Only the Surface Book 2 (with 8th gen proc) has 4 cores.  You should update your document to indicated to remove the 4 cores requirement.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5f43df7b-6ead-de5a-2e41-4d49978ab485
* Version Independent ID: 5497d300-cb21-55f3-7e80-98bc3592635c
* Content: [System requirements for Windows Defender Application Guard (Windows 10)](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-guard/reqs-wd-app-guard)
* Content Source: [windows/security/threat-protection/windows-defender-application-guard/reqs-wd-app-guard.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/windows-defender-application-guard/reqs-wd-app-guard.md)
* Product: **w10**
* GitHub Login: @Justinha
* Microsoft Alias: **justinha**

Justinha:
@tdogdfw Thanks for raising this. 
@ronsong Thanks in advance for any help you can provide.

tdogdfw:
When you are working with Intel MOBILE processors (Core-I, Pentium, Celeron and Atom) the CPU only supports 2 cores.  The new 8th generation MOBILE Core-I processors have 4 cores.  Every mobile Surface device (outside of the new Surface Book 2) is based on dual core processors since they are all 7th generation or older.  The Book2 uses an 8th generation processor which actually does have 4 cores.  This is true with all laptop/tablet devices across all OEMs which are using any of the MOBILE Intel processors prior to 8th generation Core I.  Perhaps you were querying “Logical Processors” and not cores?  Logical processors(4) = cores(2) * threads(2).

As proof, I’ve screen captured task manager from a new Core I7 Surface Pro 5 that shows only 2 cores.  All of my Surface Pro 5’s, 4’s and 3’s have been able to run hypervisor and support Application Guard, despite the comment on your documentation that says 4 cores are required.

[cid:image006.jpg@01D43569.1CEC1860]

[...]

PR #1677, opened 2018-09-24
Update reqs-wd-app-guard.md

qrscharmed:
updated with parenthesis to note cores as logical processors as well. pls let me know if it makes sense.


illfated:
Ref. #1480

PR #1706, opened 2018-09-28
Update tpm-recommendations.md

coolriggs:
Updated Exploit Guard row and the System Guard to reflect current TPM 2.0


coolriggs:
Complete

PR #1762, opened 2018-10-05
Windows 10 Lifecycle

ashishknitcs:
IT Pros need to keep tab on support for different Windows 10 Editions


jdeckerms:
@ashishknitcs Thanks for the suggestion! However, rather than duplicate information that's available elsewhere (https://support.microsoft.com/en-us/help/13853/windows-lifecycle-fact-sheet), we should link to it where appropriate. If you see a section in the content where you believe this link would be helpful, please add it.

PR #1765, opened 2018-10-05
Update windows-10-1809-removed-features.md

RonaldAi:
The feature team responsible for the TPM management console was not able to remove the console in 1809. We plan on doing this in the next release. However, because the console is not removed in 1809, we want to move the entry for the console from the "removed" section into the "no longer under development" section.


lizap:
This change is fine. Merging now.

PR #1768, opened 2018-10-05
Fix documentation for RemoteGuard command line switch for mstsc

vineelko:


vineelko:
> Shouldn't it be a "must" ?
> The user must be part of the administrators group.

Resolved it...

ChristianMontoya:
@lizap : This looks good to me. Can you also review?

PR #1861, opened 2018-10-13
Update Provisioned Apps for 1809 incl. suggested change

Borkason:
I updated the list for provisioned apps. I also want to make some suggestions to the page, which I explanatorily incorporated into the provisioned apps table in this pull request. If they are accepted, I will update the entire page accordingly:

1. Swap the first two columns *Name* and *Full name*. As a systems administrator the name of the system package is more important to us, than the canonical name.

2. Reorder the table by *Full name*. `Get-AppxProvisionedPackage` does not return a *Name*, only the *Full name*. Keeping the table up2date is currently very tedious and reordering by *Name* makes this much, much easier. I also believe that for a systems administrator, this ordering is more helpful.

3. Change the column header to *Package name* and *App name*. *Package name* is closer to the representation by the PowerShell (*DisplayName* would be exact term, but probably not the best choice in this table). And *App name* is how the app is named in the store. (E. g. Microsoft.SkypeApp <-> Skype or microsoft.windowscommunicationsapps <-> Mail and Calendar)

grafik

grafik

4. Add the links to the store page. This has been discussed in my previous pull request and supported by others. It's helpful for system administrators to be able to securely identify what package is what app. This helps when the name of the app completely deviantes from the package name, like with *microsoft.windowscommunicationsapps*, which in reality is the *Mail and Calendar* app. There are a lot of sysadmins starting with deployment, that don't know that.

### DIFF
--- a/windows/application-management/apps-in-windows-10.md
+++ b/windows/application-management/apps-in-windows-10.md
@@ -131,53 +131,58 @@ Here are the typical installed Windows apps in Windows 10 versions 1703, 1709, a
 
 ## Provisioned Windows apps
 
-Here are the typical provisioned Windows apps in Windows 10 versions 1703, 1709, and 1803.
+Here are the provisioned Windows apps in Windows 10 versions 1703, 1709, 1803 and 1809.
 
-| Name                            | Full name                              | 1703 | 1709 | 1803 | Uninstall through UI?     |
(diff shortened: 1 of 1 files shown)


Borkason:
Thanks @shortpatti 
I will soon begin applying the accepted suggestion to the rest of the page.

PR #1981, opened 2018-10-29
replace start page with Home page

aavdberg:
Replace the start page with home page because else there are used two namen for the same and this gets very confusing.


shortpatti:
The group policy values in gpedit are:

=== Detailed values: === 
Id: ConfigureHomeButtonDropdown
ValueName: ConfigureHomeButton

item: decimal: 0 => Show home button & set to Start page 

item: decimal: 1 => Show home button & set to New tab page 

item: decimal: 2 => Show home button & set a specific page 

item: decimal: 3 => Hide home button

shortpatti:
Disabled or not configured (0 default) = Load the Start page
Enabled (1) = Load the New Tab page
Enabled (2) = Load the custom URL defined in the Set Home Button URL policy
Enabled (3) = Hide the home button.

PR #2034, opened 2018-11-05
Added script to generate AppUserModelId

jaromirk:


jaromirk:
I'll create different pull...

PR #2035, opened 2018-11-05
Added AppID script

jaromirk:


jaromirk:
Ah, found it in another page …

PR #2053, opened 2018-11-08
Fix a minor typo

FrogletApps:
Replaced "Nullsoft Scriptable Install System" with "Nullsoft Sciptable Install System" in the Bondat section, and also put commas in the list to make it consistent with the other lists


jdeckerms:
@FrogletApps Thank you for fixing those items! We'll get your changes merged.

PR #2082, opened 2018-11-12
Update how-to-download-and-deploy-mdop-group-policy--admx--templates.md

CCDFischer:
add instructions on how to expand the downloaded .cab file


jdeckerms:
@CCDFischer Thanks for the change, we'll get your input merged.

PR #2087, opened 2018-11-12
Spelling correction

JohnRajunas:
Add Missing letter, changing "indows Defender ATP uses . . ." to  "Windows Defender ATP uses . . ."


lizap:
Ack - thanks for restoring that rogue W!

PR #2096, opened 2018-11-13
typo: Double word "in"

nschonni:


jdeckerms:
@nschonni I'm not going to merge this one, because the double "in" is deliberate ([logged in] in). But I'll go in and write around it.

PR #2100, opened 2018-11-13
typo: Double word "RTM"

nschonni:
Figure 2 seems to agree that is should be a single instance of "RTM"


jdeckerms:
@nschonni Thank you!

PR #2119, opened 2018-11-14
Update to add a note

kaushika-msft:
Note added to the beginning of article informing customers that they need to install the relevant ADMX files.


DaniHalfin:
Not sure how the list will render within the note section, but we'll fix it if it's weird.

PR #2137, opened 2018-11-15
Updated subscriptions

mtniehaus:
Added M365 Academic subscriptions to the list, and updated all subscriptions to include links to the subscription/product details.


jdeckerms:
@mtniehaus Thanks, Michael! Just fyi for future topics, please remove the "en-us" from Microsoft URLs -- it helps with localization. I'll fix them in this one.

PR #2150, opened 2018-11-17
Added links for Surface Laptop 2 and Surface Pro 6

emeria:
I added the links for Surface Laptop 2 and Surface Pro 6.


jdeckerms:
@emeria Thank you!

PR #2221, opened 2018-11-27

zdvorak:
Adding note about installing the latest monthly rollup


jdeckerms:
@zdvorak Thanks!

PR #2258, opened 2018-12-03
Document more information on GPOs

andreiztm:
Added more information from my internal KB on which this section was based on.


Justinha:
@andreiztm Thanks for these corrections!

andreiztm:
@Justinha thanks as well :D

PR #2278, opened 2018-12-05
Fixing small typo

anatarnousk:


Justinha:
@anatarnousk Thanks for the correction!

PR #2285, opened 2018-12-06
Update create-a-device-account-using-office-365.md

CTroessaert:
The registrar pool retrieval should be declared as a step


jdeckerms:
@CTroessaert Thanks!

PR #2311, opened 2018-12-10
CertAnalytics event changes are in-prod and must by in-doc..

MatthewMWR:


lizap:
Thanks for the fix - you'll see the changes live the afternoon of 12/14.

PR #2451, opened 2019-01-10
Fix registry key paths for HTTP.SYS

nicolonsky:
Corrected invalid registry key paths for the HTTP.SYS configuration.


illfated:
Good point, thanks for adding the missing registry folder name. 👍

PR #2453, opened 2019-01-10
typo

bb-froggy:


illfated:
Good grammar correction there, thanks. BTW, from my experience with American English computer texts, there often used to be a double space between sentences, stemming from the time when only mono-spaced typeface was used in computing. I am not sure if that is still a thing in American computing, but I find it less and less used elsewhere. I presume that double space was a typo, so it is better to have it corrected.

bb-froggy:
Ah, interesting observation about the double spaces. The other sentences are separated by only one space, so removing the double space is better, even if it is only for consistency.

PR #2466, opened 2019-01-11
Compliance architecture setps

rramoscabral:
In the Update Compliance architecture  is defined 4 steps and not five.


illfated:
Good point. Also good choice to change the step numbering format, yours looks more natural.

PR #2523, opened 2019-01-21
The upper limit of  CAPI2 log size was exteneded

tayasuta:
According to the filed engineer's request, I added the command which increases CAPI2 log default size limit.

 wevtutil sl Microsoft-Windows-CAPI2/Operational /ms:104857600


illfated:
Looks OK to me, but I don't have the required skill to evaluate any consequences involved in the change.
I can only suggest a cosmetic change, to remove 1 blank space between
```
and increase the size  
```
and the colon at the end of those 2 lines containing that text.

kaushika-msft:
@jdeckerms - Looks Okay. Please merge the changes. Thanks!

PR #2541, opened 2019-01-24
Fix typos in application-management

Nacimota:
Fixes all typos I could find in `/windows/application-management` and `/windows/application-management/app-v`.

---

The use of `SQL`-flavoured code blocks in [appv-deploy-the-appv-server-with-a-script.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/application-management/app-v/appv-deploy-the-appv-server-with-a-script.md) struck me as odd, but I left them as-is and merely removed the `<br>` tags and first sentence from the [last one](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/application-management/app-v/appv-deploy-the-appv-server-with-a-script.md#example-for-using-a-custom-instance-of-microsoft-sql-server-for-installing-the-reporting-database-on-a-different-computer-than-the-reporting-server), which I am certain should not be there (and is not consistent with the rest of the document).


illfated:
Thanks. It is a relief for nitpickers like me who get easily annoyed by repeated typos.
Furthermore, I agree with your assessment that the SQL code blocks look a bit odd, but only on the GitHub page.
Looking at the docs.microsoft.com web page result,
- https://docs.microsoft.com/nb-no/windows/application-management/app-v/appv-deploy-the-appv-server-with-a-script

it does look more normal with no syntax highlighting (only extended code block functionality).

If this topic had been part of my daily routine, I could have been more opinionated, but for now I could only think that maybe more spacing would increase readability.
Then again, seeing that the code is meant to be copied & pasted to some SQL input/query editor, I guess it could be less of an improvement and more prone to errors.

PR #2544, opened 2019-01-25
Update hello-hybrid-key-new-install.md

Ronald-den-Os:
The PowerShell command to configure AD CS should be Install-AdcsCertificationAuthority instead of Install-AdscCertificateAuthority. See https://docs.microsoft.com/en-us/powershell/module/adcsdeployment/install-adcscertificationauthority


lizap:
Thanks for catching this!

PR #2550, opened 2019-01-25
Incorrect information about deferral on Win10 1703

andreiztm:
Hi, this was tested internally by multiple engineers and for 1703 the max deferral is 365 days, the MDM CSP documentation is correct:
 Important
The default maximum number of days to defer an update has been increased from 180 (Windows 10, version 1607) to 365 in Windows 10, version 1703.
https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-update#update-deferfeatureupdatesperiodindays

Thank you,
Andrei


lizap:
@jaimeo - can you take a look here? Thanks!

jaimeo:
This change is fine--I don't think I have perms to approve it directly, but it can be merged.

PR #2577, opened 2019-01-30
Added more detail about removing Delegated Admin Permissions

brentkendall:
Added more detail about removing Delegated Admin Permissions


illfated:
Cosmetic detail, but I notice a couple of double spaces between sentences (after full stop), but it is not consistent. Is this intended or just overlooked?

brentkendall:
That was intentional (that’s how I was taught), but if it’s inconsistent with the rest of the document, you can change it to one space.

Thanks,

Brent

illfated:
Not a big deal. I presume the extra white space will be ignored, at least for certain in the `registration-auth.md` file and maybe in the resulting HTML doc page too. I was mostly curious if it was intentional.

greg-lindsay:
Merging

PR #2587, opened 2019-01-31
Fixed obvious technical errors

skycommand:
The article had some errors, including incorrect use of "\" (backslash) instead of "/" (slash), incorrect use of the word "environment", but most importantly, its incorrect use of the "feature and quality updates" phrase defeated the purpose of an entire paragraph. You see, the paragraph was trying to reassure the reader that WSUS and SCCM continue to work as intended, but these two distribute much more than just quality and feature updates. They distribute device drivers, definition updates, security updates, feature packs (e.g. .NET Framework) and service packs. The "feature and quality updates" phrase unnecessarily restricted the scope of reassurance. Wordiness is bad.

### DIFF
--- a/windows/deployment/update/fod-and-lang-packs.md
+++ b/windows/deployment/update/fod-and-lang-packs.md
@@ -16,7 +16,7 @@ ms.date: 10/18/2018
 
 As of Windows 10 version 1709, you cannot use Windows Server Update Services (WSUS) to host [Features on Demand](https://docs.microsoft.com/windows-hardware/manufacture/desktop/features-on-demand-v2--capabilities) (FOD) and language packs for Windows 10 clients locally. Instead, you can enforce a Group Policy setting that tells the clients to pull them directly from Windows Update. You can also host FOD and language packs on a network share, but starting with Windows 10 version 1809, language packs can only be installed from Windows Update.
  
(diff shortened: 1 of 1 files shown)


illfated:
I think these changes may need approval from the original author(s) of the affected text, or at least someone who are currently in charge of this page.

Cosmetic detail: I recommend consistent usage of uppercase/lowercase in abbreviations and acronyms, like `FoD` for `Features on Demand`, especially when replacing the phrase with its abbreviation.

skycommand:
@illfated 
> I think these changes may need approval from the original author(s)

Why do you think I am here then?

>  I recommend consistent usage of...

Okay.

illfated:
> @illfated
> 
> > I think these changes may need approval from the original author(s)
> 
> Why do you think I am here then?
> 
Sorry, I did not know that you are the original author. Fair enough.
I suppose I should have done more research before I wrote that.

lizap:
Thanks for the proposed changes - I made one adjustment (there's no reason to have "" around the policyh name, which is in boldface font already) and will merge the rest now.

skycommand:
@lizap Hi. Thanks for dropping by and approving the PR. About the quotation marks, however, it was for accessibility reason. Blind people cannot see boldface. Screenreader (at least those that I have worked with) seem to respect italicization but do not handle boldface properly.

PR #2636, opened 2019-02-06
Fix incorrect information about PIN complexity policy

mlindgren:
PIN complexity group policy is only supported in Computer Configuration, not User Configuration.


lizap:
@DaniHalfin - can you take a look at this?

mlindgren:
@Justinha Thanks!

PR #2665, opened 2019-02-11
task66223

v-savila:
update broken image path


Justinha:
@v-savila Thanks for the updated image!

PR #2671, opened 2019-02-12
Update self-deploying.md

mtniehaus:
Cleaned up the page, added some additional notes to address common customer pain points.


emphasismemorandum:
Default

PR #2673, opened 2019-02-12
Update windows-10-edition-upgrades.md

WK-MS:
The table on this page indicates that Win10 LTSC can be upgraded to a SAC edition without an in-place upgrade.  Changed the table to reflect that this unsupported.  Added a note that mentions an in-place upgrade is required to switch channels.


greg-lindsay:
I've mailed Wes to discuss this change. According to the mail thread I started to get details for the table, the change in this pull request is not correct.  For the time being, this request is not approved.

lizap:
Closing per feedback.

PR #2678, opened 2019-02-13
Added tips

rahulkumarsingh73690:
Added most asked question about windows 10 "how much space does windows 10 take"


greg-lindsay:
We don't link out to external sites. Only microsoft.com and other Microsoft-owned domains are allowed.

PR #2699, opened 2019-02-15
Update windows-autopilot-requirements-network.md

narkissit:
Updated the paragraph about Delivery Optimization to include new content that is now supported (Office Updates and Win32 Apps from Intune).

### DIFF
--- a/windows/deployment/windows-autopilot/windows-autopilot-requirements-network.md
+++ b/windows/deployment/windows-autopilot/windows-autopilot-requirements-network.md
@@ -41,7 +41,7 @@ In environments that have more restrictive internet access, or for those that re
 
     -   NOTE:  If Windows Update is inaccessible, the AutoPilot process will still continue.
 
--   **Delivery Optimization.**  When downloading Windows Updates, Microsoft Store apps and app updates, Office Updates and Intune Win32 Apps the Delivery Optimization service is contacted to enable peer-to-peer sharing of content, so that all devices don’t need to download it from the internet.
+-   **Delivery Optimization.**  When downloading Windows Updates, Microsoft Store apps and app updates, Office Updates and Intune Win32 Apps, the Delivery Optimization service is contacted to enable peer-to-peer sharing of content so that all devices don’t need to download it from the internet.
 
     -   <https://docs.microsoft.com/windows/deployment/update/waas-delivery-optimization>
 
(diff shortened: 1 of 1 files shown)


windows/deployment/windows-autopilot/windows-autopilot-requirements-network.md
```diff
@@ -41,11 +41,11 @@ In environments that have more restrictive internet access, or for those that re
 
     -   NOTE:  If Windows Update is inaccessible, the AutoPilot process will still continue.
 
--   **Delivery Optimization.**  When downloading Windows Updates and Microsoft Store apps and app updates (with additional content types expected in the future), the Delivery Optimization service is contacted to enable peer-to-peer sharing of content, so that all devices don’t need to download it from the internet.
+-   **Delivery Optimization.**  When downloading Windows Updates, Microsoft Store apps and app updates, Office Updates and Intune Win32 Apps the Delivery Optimization service is contacted to enable peer-to-peer sharing of content, so that all devices don’t need to download it from the internet.
```
illfated: I could be wrong, but in order for me to understand this sentence and its focus points, I would move the comma from `sharing of content, so that` back to `Office Updates and Intune Win32 Apps, the Delivery Optimization service is contacted` . In case my suggestion looks useful, here is the full line:
`-   **Delivery Optimization.**  When downloading Windows Updates, Microsoft Store apps and app updates, Office Updates and Intune Win32 Apps, the Delivery Optimization service is contacted to enable peer-to-peer sharing of content so that all devices don’t need to download it from the internet.`

narkissit: thanks! that does make more sense.

illfated:
It is quite possible that I am overthinking the end of this text line, but to my mind, it sounds like "none of the devices need to download it from the internet":
```
[...] so that all devices don’t need to download it from the internet.
```
("all devices don't need to" = "no devices need to")

Maybe it could be more precise if it says something like
"[ ... ] to enable peer-to-peer sharing of content so that **only a few** devices need to download it from the internet."

If you have got something else you think would be precise, feel free to write that instead. 😅

lizap:
Thanks, all! I edited the source to use the text @illfated suggested. You'll see the changes live today (2/19). Thanks!

PR #2720, opened 2019-02-20
Typo

Ronald-den-Os:
In the box just above Configure Company Settings is says intstalled instead of installed


lizap:
Thanks for catching!

Issue #2706, opened 2019-02-17
Typo


Hoorge:
Enterprise-managed mode also automatically redirects any browser requests tooad non-enterprise domain(s) in the container.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cb904674-d586-0a38-0e7c-7ac3e06cbcd9
* Version Independent ID: 549fbb52-6234-2a79-8604-85498e622f98
* Content: [Enable hardware-based isolation for Microsoft Edge (Windows 10)](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-guard/install-wd-app-guard)
* Content Source: [windows/security/threat-protection/windows-defender-application-guard/install-wd-app-guard.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/windows-defender-application-guard/install-wd-app-guard.md)
* Product: **w10**
* GitHub Login: @Justinha
* Microsoft Alias: **justinha**

illfated:
Thanks for pointing it out. Now, what should it say instead of "tooad", maybe "to load"?

Hoorge:
I would say "to load".

illfated:
if it had been easier to be 100% certain of what should be the correct spelling, I would open a PR to suggest the correct writing.

Justinha:
Thanks for the feedback folks. This is fixed now.

PR #2742, opened 2019-02-21
Grammar change

Jcoetsee:
A small grammar change has been made, the wording used there should be "to add" and not "to load"
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/2706


Justinha:
@Jcoetsee Thanks for the correction!

Issue #2705, opened 2019-02-17
new version is 3.6


timsmallwood:
command should be updated

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 41c2cb44-269d-3334-019a-726eb5432f1c
* Version Independent ID: 251024cc-b818-18c6-a112-4ac0276cb4b1
* Content: [Manage Windows Defender Credential Guard (Windows 10)](https://docs.microsoft.com/en-us/windows/security/identity-protection/credential-guard/credential-guard-manage#turn-off-with-hardware-readiness-tool)
* Content Source: [windows/security/identity-protection/credential-guard/credential-guard-manage.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/identity-protection/credential-guard/credential-guard-manage.md)
* Product: **w10**
* GitHub Login: @brianlic-msft
* Microsoft Alias: **justinha**

joinimran:
@timsmallwood Can you please share more details or specifics to the command which should be updated.

e0i:
Relevant changes have been proposed at PR https://github.com/MicrosoftDocs/windows-itpro-docs/pull/2750

PR #2750, opened 2019-02-22
DCG Readiness command was old

joinimran:
A new version that is 3.6 has been released and because of this, the command DG Readiness command should be updated. I have checked the module and found that the new version of DCG is 3.6 and PowerShell command has been updated to DG_Readiness_Tool_v3.6.ps1 -Disable -AutoReboot


illfated:
Looks like issue ticket #2705 can be closed now.

PR #2816, opened 2019-03-01
This page isn't needed anymore given updates to main ASR page

arottem:


jdeckerms:
@arottem thanks for the suggestion, I've passed it on to that topic's owner to consider. I'm going to close this pull request without merging it, because removing a topic from the library is a bit more complicated than simply deleting a file.

illfated:
Oh. Right. Fair point. There could be TOC files (and possibly other pages) containing links to that page, so they would also need amending.

PR #2830, opened 2019-03-05
 Formatted the raw contents and added what was missing

skycommand:
1. Unfurled the scrambled contents
2. Added the one missing item
3. Updated the description for the UserVoice forum.

My point of reference for these changes: <https://docs.microsoft.com/en-us/windows/application-management/app-v/appv-for-windows>


jdeckerms:
Thank you, @skycommand

PR #2833, opened 2019-03-05
fixed MDM OMA-URI for CommercialID

okieselbach:
fixed MDM OMA-URI for CommercialID
from Provider/ProviderID/CommercialID to the correct one Provider/MS DM Server/CommercialID


jdeckerms:
@okieselbach That value (MS DM Server) is only for Intune, but the sentence applies to generic MDM configuration. I changed the generic back and added a sentence about Intune

okieselbach:
okay this is a complete description and the best way then, I like it!

PR #2878, opened 2019-03-09
Added the most asked question about windows 10

rahulkumarsingh73690:
Added the most asked question about Windows 10 (what is the pc requirement for installing windows 10)

### DIFF
--- a/windows/whats-new/whats-new-windows-10-version-1803.md
+++ b/windows/whats-new/whats-new-windows-10-version-1803.md
@@ -232,6 +232,7 @@ Support in [Windows Defender Application Guard](#windows-defender-application-gu
 ## See Also
 
 [Windows 10 Features](https://www.microsoft.com/windows/features): Review general information about Windows 10 features.<br>
+[Windows 10 Requirements](https://goo.gl/X2ojKR): See what is the Requirements for installing Windows 10.<br>
 [What's New in Windows 10](https://docs.microsoft.com/windows/whats-new/): See what’s new in other versions of Windows 10.<br>
 [What's new in Windows 10, version 1709](https://docs.microsoft.com/windows-hardware/get-started/what-s-new-in-windows): See what’s new in Windows 10 hardware.<br>
 [Windows 10 Fall Creators Update Next Generation Security](https://www.youtube.com/watch?v=JDGMNFwyUg8): YouTube video about Windows Defender ATP in Windows 10, version 1709.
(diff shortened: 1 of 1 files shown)


rahulkumarsingh73690:
> Added the most asked question about Windows 10 (what is the pc requirement for installing windows 10)

Please Check and Merge @shortpatti

illfated:
Microsoft Docs will not accept a non-Microsoft link unless it is a partner link (and labeled as such).
Please use https://www.microsoft.com/windows/windows-10-specifications#primaryR2 instead.

rahulkumarsingh73690:
Is going through shortlink allowed or not? I am just trying to know how much clicks going throw my link..

That's all Noting Else More.

jdeckerms:
@rahulkumarsingh73690 Use the full URL (without language codes). Do not use link shorteners.

illfated:
@rahulkumarsingh73690 : Just so you know, only Microsoft internals have access to Microsoft.com domain URL shortening.
For all us others, we have to stick with full-length URLs when contributing to Microsoft Docs.
(And contributed material becomes Microsoft property, no longer your personal property.)

rahulkumarsingh73690:
Sorry. Microsoft have own URL shortner or not?

jdeckerms:
@rahulkumarsingh73690  I'm closing this pull request without merging. If you want to resubmit it, please use the full URL, without a language code.

jdeckerms:
to clarify about the language code: 

This URL is okay: https://www.microsoft.com/windows/windows-10-specifications
This URL is not: https://www.microsoft.com**/en-us**/windows/windows-10-specifications

illfated:
@rahulkumarsingh73690 : Just to clarify, regarding your question about Microsoft's own URL shortener:
Only Microsoft employees (specific ones with database access) can create Microsoft short URLs.
When needed, Microsoft will replace frequently used URLs with shortened ones.
That usually happens on official .com pages, but rarely in pure documentation pages.
For all us others, we can only use the normal URLs like mentioned above.

rahulkumarsingh73690:
Okay no problem I am just confirming about Microsoft URL Shortner.

illfated:
Not that I know of, I have not seen any publicly open and available Microsoft URL shortener yet.
I can only hope there will be one in the future, because it saves some space when editing longer lines
of documentation. especially when editing bigger chunks of text where the long URLs have a tendency
to make the text less readable during editing.

And as mentioned earlier, we can only use official Microsoft links in Microsoft Documentation pages.

egallagh:
I like that the Microsoft URL shortener has very narrow availability. I can let those go through my security. We block URLs shortened by publicly available shorteners because they are being used by bad actors to obfuscate links in malicious emails.

illfated:
Just a brainstorming idea: Would it be possible for Microsoft to configure/define/develop the URL shortener to only accept official Microsoft sites as input to keep it safe?

PR #2894, opened 2019-03-11
Fix the table

DrPizza:
Markdown tables seem very fragile. Something was upsetting this one.


illfated:
Nice job, I approve. 👍 MarkDown formatting is quite picky when it comes down to the number of spaces (or lack thereof) for text and table formatting. Following the same codestyle within the same paragraph or chapter is important too.

DrPizza:
Yeah, it looks like the single space in the line before is what was throwing it off. Easy to fix, and yet, awkward to fix, since it's so hard to see invisible characters...

illfated:
Even so, it looks like the public page https://docs.microsoft.com/windows/security/information-protection/bitlocker/bcd-settings-and-bitlocker is able to build its tables without issues, regardless of the state of the github page.

I suppose this change can be seen mostly as an improvement to keep the github page in order, just to make it easier to work on it later, if changes are needed.

DrPizza:
It looks like someone must have gone in and fixed it, because it was broken when I clicked through to edit it--showing exactly the same broken table as github's markdown....

illfated:
OK, it is possible that @Justinha has fixed it one way or other, skipping the github page.

Justinha:
@illfated @DrPizza Thanks guys. I think we might have had some temp publishing errors. Should be all fixed now.

Issue #2239, opened 2018-11-28
Incorrect Template Name 


torontojc:
The ADMX template Computer Configuration\Administrative Templates\Control Panel\Regional and Language Options\Allow input personalization is actually called Computer Configuration\Administrative Templates\Control Panel\Regional and Language Options\Allow users to enable online speech recognition services

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4662bf49-6fcd-6696-ba9d-33e573f79ea6
* Version Independent ID: b04f4b69-a99e-deb5-88ce-b4092738dead
* Content: [Use Group Policy and mobile device management (MDM) settings to configure Cortana in your organization (Windows 10)](https://docs.microsoft.com/en-us/windows/configuration/cortana-at-work/cortana-at-work-policy-settings)
* Content Source: [windows/configuration/cortana-at-work/cortana-at-work-policy-settings.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/configuration/cortana-at-work/cortana-at-work-policy-settings.md)
* Product: **w10**
* GitHub Login: @eross-msft
* Microsoft Alias: **lizross**

e0i:
@eross-msft There is a new PR #2923 available for this issue. Please review. Thanks.

eross-msft:
@e01  I am no longer part of this team. Please re-assign this to @lizap

illfated:
@officebot Assign @lizap

illfated:
@officedocsbot assign @lizap

jdeckerms:
see PR #2923

PR #2923, opened 2019-03-14
Update cortana-at-work-policy-settings.md

joinimran:
For issue https://github.com/MicrosoftDocs/windows-itpro-docs/issues/2239 the option has been updated to Allow users to enable online speech recognition services.


jdeckerms:
@joinimran I verified the change to the Group Policy setting is correct but checking both the CSP documentation and the upcoming version of Windows Configuration Designer, I see no change in the setting name so I changed that back in your pull request.

PR #2930, opened 2019-03-14
Update deploy the latest firmware and drivers for Surface devices

CoveMiner:
Updated, per CI 98622, https://vstf-us-wa-28.partners.extranet.microsoft.com:8443/tfs/ContentIdea/ContentIdea/_workitems?id=98622&triage=true&fullScreen=false&_a=edit.

PR #2932, opened 2019-03-14
Update change history

CoveMiner:
Adding System SKU reference


jdeckerms:
merge pending on merge of 2930

illfated:
Looks like #2930 is replaced by #2978 and/or #2980
(I am not certain about the content/location for those)

PR #2935, opened 2019-03-14
Update faq-wd-app-guard.md

qrscharmed:
Please do not publish until 19H1 is released as it references new 19H1 keyboards. Also, please help update the 19H1 name, I guess it's supposed to 1910?


Justinha:
@qrscharmed if we need to hold on to this for a week or more before publishing, it would be better to submit through the private repo, where we can control the release more easily.

Justinha:
this has been moved to private repo.

Issue #2284, opened 2018-12-06
BitLocker CSP with Windows 10, version 1809 Pro


banzaisi:
There is now a note at the top of this document which says this CSP is now supported in Windows 10 Pro.
However, none of the settings show as available in Windows 10 Pro (there is an "X" in each individual setting entry for Windows 10 Pro).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4a05802f-9d0f-9aeb-3c8e-8b92f1284202
* Version Independent ID: 1fd9efc2-8946-75e4-27f3-95ac43a9d987
* Content: [BitLocker CSP](https://docs.microsoft.com/en-us/windows/client-management/mdm/bitlocker-csp)
* Content Source: [windows/client-management/mdm/bitlocker-csp.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/client-management/mdm/bitlocker-csp.md)
* Product: **w10**
* GitHub Login: @MariciaAlforque
* Microsoft Alias: **maricia**

illfated:
This issue was solved in PR https://github.com/MicrosoftDocs/windows-itpro-docs/pull/2975 (commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/8675fafed6f513c9e4f8f90a1d12db7e90926c29), but those did not get referenced back to this issue page.

@banzaisi & @MarkCavins : Thank you for posting feedback regarding the lack of correct images of Windows 10 Pro being supported. As you may have observed already, 10 out of 11 policies are updated to show the correct images & support status.

([ @officedocsbot Close ])

e0i:
@officedocsbot assign @e0i

e0i:
@banzaisi 

As mentioned in the comments, this issue has been addressed with the PR #2975 and content has been updated.

We are thus closing this issue since it seems to be resolved. Feel free to re-open it if you feel it hasn't been resolved.

Thank you for taking the time to provide feedback.

e0i:
@officedocsbot close

PR #2975, opened 2019-03-19
Clarify that Pro also is supported

JankeSkanke:
It is not clear today that Pro is supported, only Business is "marked" - but the message from the Product Group is that Pro is enabled. 
I have changed the marking in all the tables to visualize this.


illfated:
Ref. #2284

PR #2978, opened 2019-03-19
Cove miner patch 99

CoveMiner:


illfated:
Suggestion for a descriptive PR title:
**Devices/Surface: new file - surface-system-sku-reference.md**

(You are of course free to choose a different title if you have got a better suggestion.)

PR #2990, opened 2019-03-20
Added additional Requirement

yvperren:
Added the Requirement that in a Proxy Environment WDAP Proxy Settings needs to be enabled and configured.


yvperren:
added Pullrequest

PR #3000, opened 2019-03-20
Update install-wd-app-guard.md

qrscharmed:
We want to add a note to users regarding the PowerShell step to ensure they understand that they need to check system requirements first. Otherwise WDAG is broken for them and they do not know why. This step impacts our WIP user testing as they aren't clear. Feel free to update with better wording.


lizap:
I updated the formatting to use the Note format (bold Note:, text in a purple box). Merging now - thanks!

qrscharmed:
Thank you!

Rona

Issue #655, opened 2018-04-04
Problem in dark mode


matifali:
Can't read tha caption of the figure in dark mode.

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 042d10ec-8f56-909c-9f66-f4c1363373b9
* Version Independent ID: 3b72e026-16ea-e9bf-a48b-abb42382ddab
* Content: [Perform an in-place upgrade to Windows 10 with MDT (Windows 10)](https://docs.microsoft.com/en-gb/windows/deployment/upgrade/upgrade-to-windows-10-with-the-microsoft-deployment-toolkit)
* Content Source: [windows/deployment/upgrade/upgrade-to-windows-10-with-the-microsoft-deployment-toolkit.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/deployment/upgrade/upgrade-to-windows-10-with-the-microsoft-deployment-toolkit.md)
* Service: **unspecified**
* GitHub Login: @mtniehaus
* Microsoft Alias: **greglin**

mypil:
@officedocsbot assign @mypil

mypil:
@matifali - are you still having this issue? Which caption & figure are you referring to? 

I tested it in my end and it works well in dark mode, I am able to read the caption on all figures.

illfated:
It is not the caption, but the individual unit names in this image:


![](https://docs.microsoft.com/en-gb/windows/deployment/images/upgrademdt-fig1-machines.png)

Figure 1. The machines used in this topic.
***
The background in the image is transparent, making the names **DC01** / **MDT01** / **PC001** melt into the black background.

mypil:
@illfated - Thanks for pointing this out. Let me see if I can get a PR created to resolve this issue.

illfated:
Looks like it may take a few more hours before the PR change migrates to the docs.microsoft.com page.
The [updated (new) file](https://raw.githubusercontent.com/MicrosoftDocs/windows-itpro-docs/master/windows/deployment/images/upgrademdt-fig1-machines.png) looks fine with black background now:
(https://raw.githubusercontent.com/MicrosoftDocs/windows-itpro-docs/master/windows/deployment/images/upgrademdt-fig1-machines.png)

mypil:
[...]

PR #3005, opened 2019-03-21
Added second row of names in white

get-itips:
to support Dark mode. https://github.com/MicrosoftDocs/windows-itpro-docs/issues/655


mypil:
@mtniehaus - can you please assist to check, approve and merge the content for this PR based on issue #655?

Thank you.

mtniehaus:
I assume the change was to make the image non-transparent?  Looks fine to me, but I'm not using dark mode.

get-itips:
@mtniehaus Hi Michael, not really, but achieve the same goal, I added the text in a white version that will only show up when enabling dark mode.

illfated:
Thank you for this solution to the issue in #655 . 👍

PR #3019, opened 2019-03-21
Removes ms.date since format was MM/DD/YYYY which was interferring.

kenwith:


illfated:
Interesting. Looks almost as if the original author of that section was in favor of European date order.

Issue #1259, opened 2018-07-10
sdbinst command line doc missing parameters


deanwiles:
Running sdbinst /? reveals command line options that are missing from this documentation, including how to install an sdb file.

Usage: sdbinst [-?] [-q] [-u] [-g] [-p] [-n[:WIN32|WIN64]] myfile.sdb | {guid} | "name"

    -? - print this help text.
    -p - Allow SDBs containing patches.
    -q - Quiet mode: prompts are auto-accepted.
    -u - Uninstall.
    -g {guid} - GUID of file (uninstall only).
    -n "name" - Internal name of file (uninstall only).

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 27fbb518-8e51-2225-b29d-4b3197a82d1d
* Version Independent ID: 2db08278-bc5a-4120-83d0-1cd56bb4aff5
* Content: [Using the Sdbinst.exe Command-Line Tool (Windows 10)](https://docs.microsoft.com/en-us/windows/deployment/planning/using-the-sdbinstexe-command-line-tool#command-line-options-for-deploying-customized-database-files)
* Content Source: [windows/deployment/planning/using-the-sdbinstexe-command-line-tool.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/deployment/planning/using-the-sdbinstexe-command-line-tool.md)
* Product: **w10**
* GitHub Login: @TrudyHa
* Microsoft Alias: **greglin**

mypil:
@officedocsbot assign @mypil

mypil:
@deanwiles - Thank you for submitting feedback.

I will get this issue over to the Win10 ITPro writing team for investigation.

Thank you for reporting and making the docs better. Much appreciated.


I made a note to request the team to update this when the work is complete.

greg-lindsay:
You're right there is an option that isn't displayed, but it is not how you would install an sdb file.  To install an sdb file you don't need to supply any options. The -q option is used however quite often so that prompts are suppressed.  The option that isn't listed here (-p) is rarely used, but it should be included for completeness.

mypil:
@deanwiles - From our understanding, the issue has been resolved but it may take a few days for the merged content to appear in the article. If you feel it hasn't been resolved, please re-open this issue.


Thank you for your contribution to make the docs better! Much appreciated!

mypil:
[...]

PR #3024, opened 2019-03-22
One parameter was missing

joinimran:
One of the command parameter -p was missing in the document. So I have updated that parameter.

Problem: https://github.com/MicrosoftDocs/windows-itpro-docs/issues/1259


mypil:
@TrudyHa - can you please assist to check, approve and merge the content for this PR based on issue #1259?

Thank you.

greg-lindsay:
What would really improve the content would be to add information that installing an SDB doesn't require any options. It looks like the person that posted the issue didn't understand that, and they seemed to think that the -p option was needed. The -p option should be documented, however so I've approved this. If I get time I'll try to edit the page and add some clarity.

PR #3052, opened 2019-03-25
Microsoft Whiteboard name

y0avb:
The name of the new whiteboard is "Microsoft Whiteboard", which will be replacing the 'old' app, "Microsoft Whiteboard 2016".
The new Whiteboard app doesn't use the user's OneDrive for Business folders any longer - maybe should revise the content?


jdeckerms:
@y0avb Per email, not merging

PR #3103, opened 2019-03-29
Follow-up from users, network requirements

maximvelichko:


jdeckerms:
@maximvelichko Just a note about Microsoft style - we don't use "click on". Where you have "Click on Open System Preferences...", it should be "Click Open System Preferences".

egallagh:
@jdeckerms Is there a current Microsoft Style Guide available?

jdeckerms:
@egallagh I don't think MSTP is available externally, but for all doc contributors, there's a public style guide at https://docs.microsoft.com/en-us/style-guide/welcome/

jdeckerms:
And actually we're moving away from "click", toward "select", since mouse-and-keyboard aren't the only input methods anymore. But changing "click" should be throughout a topic.

PR #3105, opened 2019-03-29
Changes to note

JasonJiachengZhao:
Made changes to "Take note" section to correct some information.

Removed redundant Footnote section.

### DIFF
--- a/windows/client-management/mdm/policy-csp-restrictedgroups.md
+++ b/windows/client-management/mdm/policy-csp-restrictedgroups.md
@@ -132,15 +132,7 @@ Here is an example:
 <hr/>
 
 Take note:
-* You must include the local administrator in the administrators group or the policy will fail
+* You should include the local administrator while modifying the administrators group to prevent accidental loss of access
 * Include the entire UPN after AzureAD
 
-Footnote:
-
--   1 - Added in Windows 10, version 1607.
--   2 - Added in Windows 10, version 1703.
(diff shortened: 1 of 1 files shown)


jdeckerms:
@JasonJiachengZhao Why did you remove the footnotes?

JasonJiachengZhao:
I hope this email finds you.

I attached the thread for your reference. Basically, we removed the footnote because we thought they were the result of a bad copy-paste. The schema is introduced in 1809, and we are confused why something related to 1607, 1703, 1709 and 1803 would be there.

We changed “must” to “should”, because Pranav, who implemented this, said the inclusion is recommended, but not required. In other words, the policy will not fail if you don’t include local admin. Hence the changes.

We probably have a misunderstanding of what the footnote is about. Would appreciate it if you could explain.

Thanks,
Jason

jdeckerms:
@JasonJiachengZhao The root CSP doc and the Policy CSP doc both contain tables that indicate which versions and editions a group of settings applies to. Each table contains superscript indicators, and the footnotes are the key to the superscript. I put them back in the topic (but immediately following the table since that's where they belong).

Your pull request can't be merged until you sign the contributor license agreement, thanks!

JasonJiachengZhao:
@jdeckerms Thanks for the explanation. We didn't know that the superscript exists. I signed the contributor license agreement. Would wait for it get get merged then!

Thanks,
Jason

Issue #3109, opened 2019-03-30
"AFK"


tFable:
Hello,

The text mentions: Starting with Windows 10, version 1809, Windows PE is released separately from the AFK.

Is "AFK" a typo for "ADK"?

Thanks,
t

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7ab08370-69ab-d025-cea2-c41c3aa4b908
* Version Independent ID: 9f5603f8-10e6-2651-3120-6d5627217cc9
* Content: [Prepare for deployment with MDT (Windows 10)](https://docs.microsoft.com/en-us/windows/deployment/deploy-windows-mdt/prepare-for-windows-deployment-with-mdt)
* Content Source: [windows/deployment/deploy-windows-mdt/prepare-for-windows-deployment-with-mdt.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/deployment/deploy-windows-mdt/prepare-for-windows-deployment-with-mdt.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

egallagh:
Good catch. I have submitted a pull request.
Thanks

PR #3110, opened 2019-03-30
Correct typo

egallagh:
Changed typo AFK to ADK as pointed out by user tFable.


illfated:
Ref. issue ticket "AFK" #3109

Issue #2631, opened 2019-02-06
SetupDiag Link does not work


chilleregeravi:
Setupdiag link throws a 404 Page not found error.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b3d6356b-87ff-8c3a-56ef-1270a823d3ab
* Version Independent ID: 4d6ae878-66cf-0e10-bdf3-bdd6054cbc9d
* Content: [What's new in Windows 10, version 1809](https://docs.microsoft.com/en-us/windows/whats-new/whats-new-windows-10-version-1809)
* Content Source: [windows/whats-new/whats-new-windows-10-version-1809.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/whats-new/whats-new-windows-10-version-1809.md)
* Product: **w10**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **trudyha**

Malind19:
Created PR https://github.com/MicrosoftDocs/windows-itpro-docs/pull/3135

mypil:
@officedocsbot close

PR #3135, opened 2019-04-03
SetupDiag link fixed

Malind19:
Removed .md from the link

### DIFF
--- a/windows/whats-new/whats-new-windows-10-version-1809.md
+++ b/windows/whats-new/whats-new-windows-10-version-1809.md
@@ -36,7 +36,7 @@ To learn more about Autopilot self-deploying mode and to see step-by-step instru
 
 ### SetupDiag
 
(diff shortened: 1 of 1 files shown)


mypil:
@greg-lindsay @TrudyHa - can you please assist to check, approve and merge the content for this PR based on issue #2631?

Thank you.

jdeckerms:
@Malind19 If the link isn't working, the problem is something other than the file suffix. When we link within a project (as opposed to a URL), we must include the .md suffix. In this case, I think the issue is with the relative path. Rather than change it in your pull request, I'm going to fix the link on our side (quicker to publish that way). Thanks for pointing out that the link doesn't work.

Malind19:
@jdeckerms Ok thanks. When I removed the .md suffx it worked. Same happened with another issue.

illfated:
@Malind19 : The problem seems to be that the migration process copying the Github version of the document (containing the link with the `.md` suffix) to the docs.microsoft.com site did not remove the .md suffix, even though it is supposed to do so. This error needs to be fixed by Microsoft internal technicians with access to the public pages, because the Github pages need the .md suffix.

Issue #1071, opened 2018-06-12
Just for devices/clients?


jasonjonesuk:
This article talks about "devices" but is the guidance also suitable for monitoring for potential intrusion on servers?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e5862cc7-3f36-5f2e-9ade-f7441f99f818
* Version Independent ID: b8bd1985-7b5b-ea79-976a-7d96ad69335c
* Content: [Use Windows Event Forwarding to help with intrusion detection (Windows 10)](https://docs.microsoft.com/en-us/windows/security/threat-protection/use-windows-event-forwarding-to-assist-in-intrusion-detection)
* Content Source: [windows/security/threat-protection/use-windows-event-forwarding-to-assist-in-intrusion-detection.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/use-windows-event-forwarding-to-assist-in-intrusion-detection.md)
* Product: **w10**
* GitHub Login: @tedhardyMSFT
* Microsoft Alias: **justinha**

MaratMussabekov:
Dear @jasonjonesuk ,
yes, server could also be used, as only thing needed is WinRM.
I've just created a PR to remove "Applied to: Windows 10" from the article, as it could be confusing.
Thank you

Justinha:
@MaratMussabekov Thanks for following up. It will be better to add new bullet for Windows Server than to remove Windows 10. I'll look for the PR and comment there.

mypil:
@jasonjonesuk - Thank you for submitting feedback.

From our understanding, the issue has been resolved but it may take a few days for the merged content to appear in the article. If you feel it hasn't been resolved, please re-open this issue.


Thank you for your contribution to make the docs better! Much appreciated!

mypil:
@officedocsbot close

PR #3181, opened 2019-04-05
Update use-windows-event-forwarding-to-assist-in-intrusion-detection.…

MaratMussabekov:
issue #1071
Removed reference to the supported OS, as it makes no sense (WEF could be used for any current Window OS and Windwos Server OS)

### DIFF
--- a/windows/security/threat-protection/use-windows-event-forwarding-to-assist-in-intrusion-detection.md
+++ b/windows/security/threat-protection/use-windows-event-forwarding-to-assist-in-intrusion-detection.md
@@ -13,9 +13,6 @@ ms.localizationpriority: medium
 
 # Use Windows Event Forwarding to help with intrusion detection
 
-**Applies to**
--   Windows 10
-
 Learn about an approach to collect events from devices in your organization. This article talks about events in both normal operations and when an intrusion is suspected.
 
 Windows Event Forwarding (WEF) reads any operational or administrative event log on a device in your organization and forwards the events you choose to a Windows Event Collector (WEC) server.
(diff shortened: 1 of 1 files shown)


mypil:
@tedhardyMSFT - can you please assist to check, approve and merge the content for this PR based on issue #1071?

Thank you.

MaratMussabekov:
Hello @Justinha ,
As far as I understand, this feature works not only for Windows 10 and Windows Server 2016, but also for all OS that supports WinRM, that's why I think that "Applies to" shouldn't be used here.
What do you think?

Thank you

egallagh:
If it does ally to the other versions of Windows then they should also be listed in the applies to section instead of removing the applies to section. This would create a positive affirmation of its relevance.

illfated:
I concur, much better to confirm in which versions of Windows it actually works, than not saying anything about which versions are relevant.

MaratMussabekov:
Dear @egallagh , @illfated ,
Don't you think that "Aplies to" should be for the features that works for some OS, but doesn't work for others?
As this feature works for ALL OS (at least since Windows Vista and Windows Server 2008), should we specify all OS there? What do you think?
Thank you

illfated:
Well, you sort of have got a point there. Speaking for myself, I would (more often than not) expect from a Microsoft page to state in which OS **versions** the documented program or software package is supported (or is expected to work). If the versions are not listed, I would expect that the documented software would be available in _all_ versions of Windows (or the software environment in question).

I could very well be mistaken, so I refer to the Microsoft Docs team members for verification and proper feedback.

MaratMussabekov:
Dear @Justinha , added Windows Server, as you recommended, could you please have a look?

Thank you

Justinha:
@MaratMussabekov Thanks for the addition!

PR #3170, opened 2019-04-05
Grammar and typo corrections in enable-attack-surface-reduction.md

illfated:
**Proposed changes:**
- conjuction => conjunction
- , it's owned by =>  is owned by
- microsoft => Microsoft
- admins => administrators
- Microsoft CLoud's Protection => Microsoft Cloud Protection
- which rules or exclusions apply to. => which rules or exclusions apply to them.

Please review and comment, in particular to request any changes, so I can update this PR directly.

Ref. https://github.com/MicrosoftDocs/windows-itpro-docs/issues/1880#issuecomment-479734075

PR #3219, opened 2019-04-09
Minor typo correction in enable-attack-surface-reduction.md

illfated:
**Change proposed:**
- `conjuction` => conjunction

Ref. #3170 (previous attempt)

BTW: Would it be useful to change "admins" to "local administrators" in the paragraph shown below, or is the current wording enough to cover the factual meaning as is?
> The rule **Block executable files from running unless they meet a prevalence, age, or trusted list criterion** with GUID 01443614-cd74-433a-b99e-2ecdc07bfc25 is owned by Microsoft and is not specified by admins.


mypil:
@AndreaBarr - Can you possibly merge the content of this PR? This is just a simple typo issue. Thank you.

PR #3229, opened 2019-04-10
Update hello-hybrid-cert-trust-prereqs.md

secherka:
To avoid customers and support engineers confusion I would recommend following change in the mentioned part of the documentation (replace first two sentences):
Windows Hello for Business hybrid certificate trust requires Active Directory being federated with Azure Active Directory and needs Windows Server 2016 Active Directory Federation Services or newer. Windows Hello for Business hybrid certificate trust doesn’t support Managed Azure Active Directory using Pass-through authentication or password hash sync.

### DIFF
--- a/windows/security/identity-protection/hello-for-business/hello-hybrid-cert-trust-prereqs.md
+++ b/windows/security/identity-protection/hello-for-business/hello-hybrid-cert-trust-prereqs.md
@@ -82,7 +82,7 @@ Organizations using older directory synchronization technology, such as DirSync 
 <br>
 
 ## Federation ##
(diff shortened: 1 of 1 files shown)


mapalko:
@DaniHalfin to review the change.

DaniHalfin:
I think there's a middle ground here. The inital text does some explaning that provides context, before going into requirement and what's not supported. I would like to keep that first part in for context and state requirements and unsuppored scenario more clearly (as done in the change) after.
@secherka

DaniHalfin:
and also, thanks for taking the time to contribute here!!

secherka:
If I understood correct, you want to keep the first sentence for general education and just replace second sentence with proposed two sentences?

Issue #2388, opened 2018-12-28
Error in example


JGezels:
Hi

It seems there is an error in the following example:

Exclude all local users.                                  /ue:%computername%*

This should be: /ue:%computername%\*            =&gt; Missing backslash

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7c7e6db2-1517-e9b9-cd9a-6a6aaf84ccc3
* Version Independent ID: daab80f9-640e-36c9-e60a-084901b11c05
* Content: [ScanState Syntax (Windows 10)](https://docs.microsoft.com/en-us/windows/deployment/usmt/usmt-scanstate-syntax#bkmk-useroptions)
* Content Source: [windows/deployment/usmt/usmt-scanstate-syntax.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/deployment/usmt/usmt-scanstate-syntax.md)
* Product: **w10**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

illfated:
Thank you for pointing out the missing backslashes (either replaced by `<em>` or simply missing). Sorry for the delay in providing a solution. I will check for and provide a solution if nobody else have made it yet.

PR #3261, opened 2019-04-15
Corrected malformed formatting tag

isjamesalive:
Replaced an unescaped backslash with HTML code to prevent formatting error.


illfated:
Cosmetic detail, but perhaps change lowercase `c:\` to uppercase `C:\` while you are there?

illfated:
Nice. That fix is actually needed, seeing the original page. You might want to include link(s) to the targeted docs.microsoft.com page in future PRs :
- https://docs.microsoft.com/en-us/windows/deployment/usmt/usmt-scanstate-syntax#a-href-idbkmk-storageoptionsastorage-options

Maybe you also know how to fix issue #2388 as well, since it is linked from the same page?

isjamesalive:
Noted, thank you.

I'll take a look at that issue too.

PR #3264, opened 2019-04-15
Update windows-analytics-azure-portal.md

manuelhauch:
Abbreviation for Azure Active Directory is AAD


jdeckerms:
@manuelhauch Actually, the approved abbreviation for Azure Active Directory is Azure AD -- not AAD. I fixed your change.

PR #3267, opened 2019-04-15
Update microsoft-recommended-block-rules.md

wweibull:
Added two additional researcher to the acknowledgment section.

Philip Tsukerman (@PhilipTsukerman)
Jimmy Bayne (@bohops)


Justinha:
@wweibull thanks for adding these finders!

PR #3319, opened 2019-04-21
Fix typo in one header

karammasri:
Fix typo in the header for "Hybrid Azure AD joined provisioning in a Key Trust deployment in a Managed environment"

Link at the top of the article to that header was not working due to the typo.


illfated:
This fix works for the link in the main article text and is also a fair typo correction.
So far, so good. Then there is the section "**In this article**" on the right side of the public page
https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-how-it-works-provisioning

In that section, the 3rd link contains the same typo as the heading, so it works.
Then again, is it simply collecting the linked article headings?
If yes, I presume this PR solution will migrate dynamically.

karammasri:
> This fix works for the link in the main article text and is also a fair typo correction.
> So far, so good. Then there is the section "**In this article**" on the right side of the public page
> https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-how-it-works-provisioning
> 
> In that section, the 3rd link contains the same typo as the heading, so it works.
> Then again, is it simply collecting the linked article headings?
> If yes, I presume this PR solution will migrate dynamically.

The link on the right side is built automatically and should be fixed once the PR is approved.

Issue #2789, opened 2019-02-27
Link issue 


danjpadgett:
The hyperlink for this text "Hybrid Azure AD joined provisioning in a synchronous Certificate Trust deployment in a Federated environment"

does not go anywhere. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7f170bfe-711d-f875-8d5f-a49d3d5113b2
* Version Independent ID: a75c9aa8-b47b-51dd-9b84-c63bb5c6b782
* Content: [How Windows Hello for Business works - Provisioning](https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-how-it-works-provisioning#Hybrid-Azure-AD-joined-provisioning-in-a-synchronous-Certificate-Trust-deployment-in-a-Federated-environment)
* Content Source: [windows/security/identity-protection/hello-for-business/hello-how-it-works-provisioning.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/identity-protection/hello-for-business/hello-how-it-works-provisioning.md)
* Product: **w10**
* GitHub Login: @MikeStephens-MS
* Microsoft Alias: **mstephen**

illfated:
Now this is weird. For me, all the internal page (anchor) links work in the Github page
(https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/identity-protection/hello-for-business/hello-how-it-works-provisioning.md),
but none of the page anchor links work in the public page
(https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-how-it-works-provisioning).

I have tested this using Microsoft Edge (Windows 10 Home 64-bit, Version 1809 (OS Build 17763.316)), Firefox 65.0.1 (64-bit) and Google Chrome Version 72.0.3626.119 (Official Build) (64-bit).
I could accept that Firefox & Chrome might block the linking on the public page due to ad-block addons, but I would expect that the links would work at least on Microsoft Edge without addons.

All the browsers can use the links on the Github page, though. Weird. I don't know why.

Clicking the anchor symbol after the chapter headlines works as expected: it adds the correct link in the address bar and makes that headline top of the visible frame of my browsers.

illfated:
Oh. I think I may have found out why the URL text links don't work on the public site.
[...]

PR #3320, opened 2019-04-21
fix links

karammasri:
The link targets at the top of the article were using the wrong casing, which caused the links to be invalid and not work.

The links were fixed to be all lowercase as required.


illfated:
Same type of issue as #2789 (**Link issue**) and the solution in commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/fc7cf615714e9a4a439678eb2e2fbe8a07cfb2ee#diff-b6eff0fef3b4c921658f88f045dfb84e .

Looks good to me™, I approve. 👍

PR #3330, opened 2019-04-22
Update credential-guard-requirements.md

cchapin2020:
Microsoft Virtual Academy is being site is being shutdown 4/30/19


illfated:
Looks good to me™.

PR #3331, opened 2019-04-22
Update credential-guard-how-it-works.md

cchapin2020:
Microsoft Virtual Academy site is being shut down 4/30/19, removed link to video


illfated:
Looks good to me™.

PR #3332, opened 2019-04-22
Md branch3122299

ManikaDhiman:
Hi Nick,

Please review the changes in this PR and provide feedback.

Thanks,
Manika


nickbrower:
Hi Manika,

One thing I’m noticing as I scroll down the diff of this PR is the order in which you document the footnote before or after the lists for additional SKUs (HoloLens, SurfaceHub, etc.).

For instance if you look at your Accounts policies<https://github.com/MicrosoftDocs/windows-itpro-docs/blob/b618ef0fb5afa21d451cbb1ee7a04170fa81728f/windows/client-management/mdm/policy-csp-accounts.md> page you will see the order is:


  1.  
  2.  <--StartHoloLens--> . . . <--EndHoloLens-->
  3.  Footnotes:

However, if you look at that same page before your edits<https://raw.githubusercontent.com/MicrosoftDocs/windows-itpro-docs/master/windows/client-management/mdm/policy-csp-accounts.md> there is no HoloLens list but the order is still different:


  1.  Footnote:
  2.  

The reason this is important is because my automation scripts expect a consistent order to parse the different area name pages.  Currently I’m expecting this order:


  1.  Footnote:
  2.  
  3.  Policy lists for additional SKUs

By changing the order you are effectively breaking my ability to parse the content in these pages in a consistent way.  I’m definitely okay with changing the order if you think it looks better that way.  It is a trivial fix on my side, but that means the order needs to be changed on all the other area name files so that it is consistent.

Here are examples of two inconsistent schemas currently checked into your branch that would currently break my parser:


  *   Accounts policies<https://raw.githubusercontent.com/MicrosoftDocs/windows-itpro-docs/b618ef0fb5afa21d451cbb1ee7a04170fa81728f/windows/client-management/mdm/policy-csp-accounts.md>:
     *   
     *   <--StartHoloLens--> . . . <--EndHoloLens-->
     *   Footnotes:
  *   AboveLock policies<https://raw.githubusercontent.com/ManikaDhiman/windows-itpro-docs/MDBranch3122299/windows/client-management/mdm/policy-csp-abovelock.md>:
     *   Footnote:
     *   

Essentially I cannot parse the two files above with a single schema. If you are to change the order in Accounts, then I also need you to change it in AboveLock.  I’ll see if I can get my validation script to work on your branch, this way I can generate a report for you of inconsistencies I find (for instance a policy that exists in a list on its area name file (Accounts or AboveLock above), but not found in the corresponding index file’s list).  Currently those inconsistencies would be found for all Accounts policies in the HoloLens list because my parser technically wouldn’t find any list at all.

I hope this makes sense, let me know if you have any questions.

-Nick

illfated:
New PR: #3359

Issue #3268, opened 2019-04-15
powershell, EnableFastFirstSignin, $namespaceName


jszabo98:
How do you set EnableFastFirstSignin in powershell?  By the way, the $namespaceName variable is undefined.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 14b376e2-39bd-faca-a76c-dd21bad8d8fd
* Version Independent ID: a76cabb0-59b9-7e97-6603-eec7d37f1fd4
* Content: [Set up a shared or guest PC with Windows 10 (Windows 10)](https://docs.microsoft.com/en-us/windows/configuration/set-up-shared-or-guest-pc)
* Content Source: [windows/configuration/set-up-shared-or-guest-pc.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/configuration/set-up-shared-or-guest-pc.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jdeckerMS
* Microsoft Alias: **jdecker**

mypil:
@officedocsbot assign @mypil

get-itips:
Hello @jszabo98 thanks for your feedback. I created PR to correct the Ps script that is present on the article, I appreciate. About the other question, currently, I encourage you to look at this resource https://docs.microsoft.com/en-us/windows/client-management/mdm/index to check for supported procedures.
Thanks!

mypil:
@jszabo98 - From our understanding, the issue has been resolved but it may take a few days for the merged content to appear in the article. If you feel it hasn't been resolved, please re-open this issue.

Thank you for your contribution to make the docs better! Much appreciated!

mypil:
@officedocsbot close

PR #3393, opened 2019-04-29
Corrected sharedPC example

get-itips:
$namespaceName variable was undefined, was replaced with the actual Namespace string

https://github.com/MicrosoftDocs/windows-itpro-docs/issues/3268

@mypil


mypil:
@jdeckerms - Can you please assist to check, approve and merge the content for this PR based on issue #3268?

Thank you.

greg-lindsay:
This is a correct edit. We could also just change the entire last line to $sharedPC which would run the same command.

Issue #3379, opened 2019-04-27
This sentence needs some clarification...


jwojan:
"This policy is used to ensure that MDM policy wins over GP when same setting is set by both GP and MDM channel. The default value is 0. The MDM policies in Policy CSP will behave as described if this policy value is set 1. "

So is it saying that a value of Zero produces the same result as if the value were set to 1? Or is it trying (and failing to say) that when at zero, GPO wins over Intune Policy?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: becd9bd1-6396-3f4a-cc78-c9a06b781516
* Version Independent ID: efeec2a0-08b0-d1fc-72c5-14c23fa11397
* Content: [Policy CSP - ControlPolicyConflict](https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-controlpolicyconflict#controlpolicyconflict-mdmwinsovergp)
* Content Source: [windows/client-management/mdm/policy-csp-controlpolicyconflict.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/client-management/mdm/policy-csp-controlpolicyconflict.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @MariciaAlforque
* Microsoft Alias: **maricia**

mypil:
@officedocsbot assign @mypil

mypil:
@jwojan - Thank you for submitting feedback.

I will get this issue over to the Win10 ITPro writing team for investigation.

Thank you for reporting and making the docs better. Much appreciated.


I made a note to request the team to update this when the work is complete.

joinimran:
@jwojan ,Thank you for sharing your query. Please note that this policy says that MDM policy will win over GP. If you have set up this policy using GP and then configured it through MDM, the MDM will have high priority and will be executed. Hope this answers your question.

Thanks.

jwojan:
I knew the answer. The published sentence makes it sound like MDM wins whether the value is set or not. That is the part that needs clarification.

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

________________________________

mypil:
@jwojan - From our understanding, the issue has been resolved but it may take a few days for the merged content to appear in the article. If you feel it hasn't been resolved, please re-open this issue.

Thank you for your contribution to make the docs better! Much appreciated!

[...]

PR #3598, opened 2019-05-09
Sentence was confusing

joinimran:
The sentence was confusing so I made a correction where now it makes more sense that when MDM policy is configured, it will win over GP.

Problem: https://github.com/MicrosoftDocs/windows-itpro-docs/issues/3379


mypil:
@Justinha @DaniHalfin - would you be able to help merge the content of this PR based on issue #3379? Or can you help refer us to the new author handling this article? 

Thank you.

PR #3609, opened 2019-05-09
Updating Domain/ComputerName node info

msft-jinayoon:
Incorporating the following info:

•	This node edits the DNS hostname of the computer.
•	If using the %RAND:x% or %SERIAL% macros, the new name is limited to 15 characters
•	If the serial number generated from the %SERIAL% macro is too long, the serial number will be truncated from the beginning of the serial number sequence, not the end. (e.g. 123ABCDEF456 --> CDEF456)
•	If the new name is a constant string (i.e. not using any of the macros) the new name can be up to 63 characters long
•	Validation for accepted characters are based on the SetComputerNameEx function: https://docs.microsoft.com/en-us/windows/desktop/api/sysinfoapi/nf-sysinfoapi-setcomputernameexa 
•	This node does not work properly for hybrid-joined AAD/AD devices (it only works for fully AAD joined devices)

I would love feedback and additional edits on how to make the proposed changes more customer-doc-friendly. Thanks!

### DIFF
--- a/windows/client-management/mdm/accounts-csp.md
+++ b/windows/client-management/mdm/accounts-csp.md
@@ -26,9 +26,13 @@ Root node.
 Interior node for the account domain information.
 
 <a href="" id="domain-computername"></a>**Domain/ComputerName**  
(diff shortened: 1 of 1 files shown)


emphasismemorandum:
Fraud

Issue #874, opened 2018-05-08
1803: Removal / De-Provisioning of Apps is broken, not longer functional


gruppenrichtlinien:
1803 specific:
Get-AppxProvisionedPackage -Online | Where-Object {$_.Displayname -Like "Microsoft.BingNews"} | ForEach-Object { Remove-AppxProvisionedPackage -Online -PackageName $_.PackageName}
or alternative: Get-AppxProvisionedPackage -Online | where {$_.DisplayName -eq "Name.of.The.App"} |
Remove-AppxProvisionedPackage -online
does not work in 1803. Apps will still be installed for each new user.

DISM can do it: cmd /c DISM.exe /online /Remove-ProvisionedAppxPackage /PackageName:"Package.Name"
but dism will also delete the ressources in %programfiles%\WindowsApps and the App can not be restored

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e8228962-7158-6be1-e791-e6535b88f343
* Version Independent ID: 46e4c7f3-e8b0-ebc9-da58-b69149803937
* Content: [Manage connections from Windows operating system components to Microsoft services (Windows 10)](https://docs.microsoft.com/en-us/windows/configuration/manage-connections-from-windows-operating-system-components-to-microsoft-services)
* Content Source: [windows/configuration/manage-connections-from-windows-operating-system-components-to-microsoft-services.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/configuration/manage-connections-from-windows-operating-system-components-to-microsoft-services.md)
* Service: **w10**
* GitHub Login: @brianlic-msft
* Microsoft Alias: **brianlic-msft**

heine-it-yh:
I can confirm this, Remove-AppxProvisionedPackage throws an unspecified error "COMException".

Reproducible with
```
$ProvisionedPackage = Get-AppxProvisionedPackage -online | Where-Object {$_.displayName -CMatch "app name as regex"}
remove-AppxProvisionedPackage -online -LogLevel 2 -packagename $ProvisionedPackage.PackageName
```

Mack-IT-CA:
I also attempted the DISM command referenced earlier and received an error:

[...]

PR #3626, opened 2019-05-11
Important note for #874

j0rt3g4:
Added important note about how to resolve that fact that the cmdl doesn't work. #874


windows/privacy/manage-connections-from-windows-operating-system-components-to-microsoft-services.md
```diff
@@ -768,7 +768,9 @@ To remove the News app:
 - Right-click the app in Start, and then click **Uninstall**.
 
   -or-
-
+>[!IMPORTANT]
+> If you have any issue with this commands, go ahead a do a system reboot,and try the scripts again.
```
illfated: You need to either use **`these commands`** (plural) or **`this command`** (singular)
(I don't know if the note refers to many commands or just one command.)
("this commands" turns "commands" into a verb, like in "This commands us to do something else.)

j0rt3g4: Agree lol it's plural there

windows/privacy/manage-connections-from-windows-operating-system-components-to-microsoft-services.md
```diff
@@ -768,7 +768,9 @@ To remove the News app:
 - Right-click the app in Start, and then click **Uninstall**.
 
   -or-
-
+>[!IMPORTANT]
+> If you have any issue with these commands, go ahead a do a system reboot, and try the scripts again.
```
illfated: After re-reading the text (after the correction), I think it will be better to simplify the sentence like this:
```suggestion
> If you have any issues with these commands, do a system reboot and try the scripts again.
```

lizap:
Merging as is, but in the future, can you use "restart the PC" or even "restart the system," instead of "do a system reboot"? Thanks!

j0rt3g4:
@lizap I can correct it np

j0rt3g4:
#3706 @lizap Here's the correction.
Thank you

PR #3690, opened 2019-05-15
Multiple Typos

nenonix:
Typos fixed, "prerequistes" and a few others


nenonix:
@lizap I did not include typos in metadata in that docset,  Will make a PR for those too. Thanks

PR #3700, opened 2019-05-15
Changed to 26.1 instead of 27.1 after 26.

EvansShaun:
Followed link later in documentation to confirm that was the desired numbering.


illfated:
There are 2 more `27.1 Apps for websites` in the tables on the same page, but only 1 of those is followed by 27 (the other one is followed by 28). You may want to add those to your PR content.

EvansShaun:
Thank you.  I fixed those two additional 26.1, vs 27.1.  There is a different format where some sections will skip something altogether, but the bookmarks match something specifically later in the documentation.

illfated:
Very well, fair enough. Good point, I was uncertain about the last incorrect table link, where the actual ordinal number 27 was missing between 26 and 28. Happy to see that you have got it covered. 👍

PR #3721, opened 2019-05-16
Update windows-autopilot-requirements.md

Mizanurra:
This is a note for IT admins to understand the Diagnostics data collection settings requirements for Autopilot deployed devices.
As part of Windows 19H1 release OOBE default Diagnostic Data collection settings would be set to Basic but for Autopilot Deployed devices OOBE would set Diagnostics Data collection settings to Full.

### DIFF
--- a/windows/deployment/windows-autopilot/windows-autopilot-requirements.md
+++ b/windows/deployment/windows-autopilot/windows-autopilot-requirements.md
@@ -49,8 +49,7 @@ See the following topics for details on network and configuration requirements:
 There are no additional hardware requirements to use Windows 10 Autopilot, beyond the [requirements to run Windows 10](https://www.microsoft.com/windows/windows-10-specifications).
 
 >[!IMPORTANT]
-   >It is a notice of awareness to the administrator that’s configuring Windows Autopilot on Windows 10, version 1903 and onwards.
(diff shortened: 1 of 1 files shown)


e0i:
@DaniHalfin Would you be able to review and merge this PR opened by a MS contributor? Thanks.

Mizanurra:
> This is not a requirement, but rather a fact about the device state during OOBE. We shouldn't publish this to the requirements page. Also I would say it is just a note or tip, not an important hint.

@greg-lindsay  Hi Greg, this is guidance we got from CELA and Privacy that we should add this comments as Autopilot requirements. Some background, starting 1903 Windows 10 is going to set diagnostics telemetry settings to BASIC by default which would break Autopilot deployment state for IT. Starting 1903 release Autopilot client would enforce Diagnostic settings to FULL for Autopilot devices.  By virtue of using Autopilot and hiding Privacy settings from OOBE IT must understand that using Autopilot would enforce diagnostics telemetry settings to be FULL instead of BASIC. Hope this help.

Mizanurra:
> This is not a requirement, but rather a fact about the device state during OOBE. We shouldn't publish this to the requirements page. Also I would say it is just a note or tip, not an important hint.
Hi @greg-lindsay 
I just noticed that we have new table under network requirements which has multiple services and there is a line item with "Diagnostics  Data" We should update that messaging with something similar or combine them into one and convey that for Autopilot deployed devices Diagnostics settings is defaulted to FULL during OOBE.

PR #3749, opened 2019-05-20
Asymmetrical is not a word

SteveSyfuhs:


illfated:
I agree with the content change, because the main difference between _asymmetric_ and _asymmetrical_ is the functionality and how to use them. I am quite certain that asymmetrical is a word, it simply comes down to usage. On that basis, I only disagree with the conclusion in the PR title.

Issue #3805, opened 2019-05-24
[LOC]"&lt;"needs to be replaced with"<", "&gt;"needs to be replaced with">"


hyoshioka0128:
https://docs.microsoft.com/ja-jp/windows/application-management/manage-windows-mixed-reality
![manage-windows-mixed-reality](https://user-images.githubusercontent.com/40815708/58345284-ec856180-7e92-11e9-8f59-61c1f21cd538.png)

---
#### ドキュメントの詳細

⚠ *このセクションを編集しないでください。 docs.microsoft.com で必須です ➟ GitHub の問題のリンク。*

* ID: 98d4c6ca-ae3c-7750-f58b-0df996e41402
* Version Independent ID: 0d05fd22-b761-6fb4-8686-a2021573e542
* Content: [企業で Windows Mixed Reality アプリを有効化またはブロックする (Windows 10)](https://docs.microsoft.com/ja-jp/windows/application-management/manage-windows-mixed-reality)
* Content Source: [windows/application-management/manage-windows-mixed-reality.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/application-management/manage-windows-mixed-reality.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jdeckerMS
* Microsoft Alias: **jdecker**

illfated:
Future solution: PR #3807 -- when merged into the `master` branch and migrated to the Docs page, subsequently migrated to the localized page, it should close this issue.

(Comments are welcome regarding estimated time of total time before arrival on the localized pages.)

e0i:
@officedocsbot assign @e0i

e0i:
@officedocsbot assign @e0i
@officedocsbot assign @dongmicro
@officedocsbot add label localization

hyoshioka0128:
@illfated : #3807 (ja-jp) was also fixed.

illfated:
Nice! The PR changes have migrated both to the docs.microsoft.com/en-us/ page and to the /ja-jp/ page. Very good indeed.

@e0i : This issue ticket is now resolved and can be closed. Ref. the XML content in

[...]

PR #3807, opened 2019-05-24
Typo "&lt;"→"<", "&gt;"→">"

hyoshioka0128:
https://docs.microsoft.com/en-us/windows/application-management/manage-windows-mixed-reality

### DIFF
--- a/windows/application-management/manage-windows-mixed-reality.md
+++ b/windows/application-management/manage-windows-mixed-reality.md
@@ -72,22 +72,22 @@ In the following example, the **Id** can be any generated GUID and the **Name** 
                     <Type xmlns="syncml:metinf">text/plain</Type>
                 </Meta>
                 <Data>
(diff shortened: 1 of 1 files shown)


illfated:
Definitely a good fix, looks good to me. I approve.
@DaniHalfin , do you have time to verify this PR or check if there are someone else who would like to confirm this solution?

When merged and then migrates to the localized pages (whenever that will be), it will also close issue ticket #3805 .

illfated:
Looks like the offending format ( `&lt;` / `&gt;` ) has been unchanged since the material was added in commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/82f3d9d64be674ccb73d9a7984f018d46da92af8 (Oct 10, 2017).

e0i:
@DaniHalfin The PR relates to a localization issue that has been escalated to the localization team. Would you approve suggested changes anyway in the meantime the localization fixes take effect? Thank you.

PR #3873, opened 2019-05-31
fix typo avilable

Styxxy:


illfated:
Nice catch. Thank you. I tend to get distracted too easily by typos like that.

Issue #3819, opened 2019-05-27
[LOC]"&lt;"needs to be replaced with"<", "&gt;"needs to be replaced with">"


hyoshioka0128:
https://docs.microsoft.com/ja-jp/windows/configuration/kiosk-mdm-bridge
![kiosk-mdm-bridge](https://user-images.githubusercontent.com/40815708/58407092-f206c580-80a5-11e9-9dbb-3bf6272faa3f.png)

---
#### ドキュメントの詳細

⚠ *このセクションを編集しないでください。 docs.microsoft.com で必須です ➟ GitHub の問題のリンク。*

* ID: 7f536979-dfd7-317d-2da7-c40935d85c7f
* Version Independent ID: c23737a1-c71e-3d9c-3915-c6929a23602e
* Content: [MDM ブリッジ WMI プロバイダーを使用して、Windows 10 キオスク (Windows 10) を作成するには](https://docs.microsoft.com/ja-jp/windows/configuration/kiosk-mdm-bridge)
* Content Source: [windows/configuration/kiosk-mdm-bridge.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/configuration/kiosk-mdm-bridge.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jdeckerMS
* Microsoft Alias: **jdecker**

e0i:
@officedocsbot assign @e0i

illfated:
I wonder if the code content has been created in some form of HTML editor, converting any special characters like ` < `, ` " ` and ` > ` to shorthand names for those codes.
Would it be reasonable to presume that the content should look like this, with the modified characters reverted back like this:
```xml
$nameSpaceName="root\cimv2\mdm\dmmap"
$className="MDM_AssignedAccess"
$obj = Get-CimInstance -Namespace $namespaceName -ClassName $className
$obj.Configuration = @"
<?xml version="1.0" encoding="utf-8" ?>
<AssignedAccessConfiguration xmlns="http://schemas.microsoft.com/AssignedAccess/2017/config">
  <Profiles>
    <Profile Id="{9A2A490F-10F6-4764-974A-43B19E722C23}">
      <AllAppsList>
        <AllowedApps>
          <App AppUserModelId="Microsoft.ZuneMusic_8wekyb3d8bbwe!Microsoft.ZuneMusic" />
[...]

PR #3955, opened 2019-06-08
Windows/Configuration: correct HTML code to XML tags

illfated:
**Description:**

The XML code block contains HTML code instead of the expected XML tags.

**Changes proposed :**

Converting these to proper XML format should at least make it readable.
The intended outcome of using correct XML tags is to make it work as expected.
Also replacing "ps" with "xml" as the correct code block syntax name for code highlighting.

**Issue ticket to be closed or referenced:**

Ref. #3819 **[LOC]`"&lt;"`needs to be replaced with"<", `"&gt;"` needs to be replaced with">"**

(Should close issue ticket #3819 after migration to docs.microsoft.com/en-us/ and later updated to the Japanese page.)

### DIFF
--- a/windows/configuration/kiosk-mdm-bridge.md
+++ b/windows/configuration/kiosk-mdm-bridge.md
@@ -31,59 +31,59 @@ Here’s an example to set AssignedAccess configuration:
 3. In the command prompt launched by psexec.exe, enter `powershell.exe` to open PowerShell. 
 4. Execute the following script:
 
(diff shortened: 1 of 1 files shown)


e0i:
@dansimp Would you be able to review and merge this PR contributed by the community? The issue is originally passed to localization team for review and PR seems to intend to bring a solution in the meantime. Thank you.

PR #4044, opened 2019-06-16
chore: Add passing Markdownlint config

nschonni:
Suppressed all failing rules so editing in VS Code has less noise


e0i:
@lizap @danihalfin @Dansimp Would you be able to review this PR that introduces a VS Code configuration? Thanks.

.markdownlint.json
```diff
@@ -0,0 +1,39 @@
...
+    "MD029": false,
+    "MD030": false,
+    "MD031": false,
+    "MD032": false,
+    "MD033": false,
+    "MD034": false,
+    "MD036": false,
+    "MD037": false,
+    "MD038": false,
+    "MD039": false,
```
nschonni: Should be removed since #4045 landed

.markdownlint.json
```diff
@@ -0,0 +1,39 @@
...
+    "MD040": false,
+    "MD041": false,
+    "MD042": false,
+    "MD045": false,
```
nschonni: Should be removed since #4046 landed

e0i:
@lizap @DaniHalfin @Dansimp Any updates with this PR? Thanks.

DaniHalfin:
I'd rather keep any unnecessary config files out of the repo.

nschonni:
@DaniHalfin if you're editing articles with VS Code and the MS Docs Authoring Pack, many articles have warnings throughout without adding this

DaniHalfin:
Agreed, but this is a tooling config and pertains to the preferences of an individual. I don't want to push (or maintain, for that matter) that kind of configuration.

nschonni:
The current PR isn't a preference, it just silences the currently failing rules so they can be cleaned up later. Some teams do actually go the preference route though like https://github.com/aspnet/AspNetCore.Docs/blob/master/.markdownlint.json

DaniHalfin:
The preference part I was talking about, is whether or not to use markdown lint (even with the authoring pack installed, markdwon lint can be disabled). 
Adding and maintaining tooling configuration in a content repo isn't something we're looking to do in our content repo at this point, but we can revisit if we get more feedback to that point.

e0i:
@nschonni 

From our understanding the pull request has been declined via **DaniHalfin**'s comments and thus it is being closed.

Thank you so far with the rest of your contributions. Much appreciated.

e0i:
@officedocsbot close

PR #4049, opened 2019-06-17
chore: Renormalize line endings

nschonni:
Ran `git add --renormalize .`

PR #4060, opened 2019-06-17
typo: Double word "be"

nschonni:


illfated:
I presume the change(s) in line 91 through 94 might be that one or more invisible characters have been replaced by a similar invisible character from a different character set.

nschonni:
Looks like it is one that got affected in #4049 so it probably has mixed line endings

e0i:
See #4062

illfated:
Heh, I missed out on making a joke about "Eric the half a Bee" + "To be or not to be" there. Oh well. Good job, Nick.

PR #4065, opened 2019-06-17
Updates considered to be out-of-date

manuelhauch:
The description in the Group Policy Editor says: "By default, this value is set to 14 days." for both: "Define the number of days before spyware security intelligence is considered out of date" and "Define the number of days before virus security intelligence is considered out of date".

The Link refers to those two settings.


e0i:
@Dansimp This PR that introduces a minor improvement is ready for review. Thanks.

PR #4098, opened 2019-06-19
typo: oganization -> organization

nschonni:

PR #4101, opened 2019-06-19
chore: Remove link locale

nschonni:


e0i:
See: #4098

windows/deployment/upgrade/setupdiag.md
```diff
@@ -404,9 +404,9 @@ Refer to https://docs.microsoft.com/windows/deployment/upgrade/upgrade-error-cod
 
 ### XML log sample
 
-```
+```xml
 <?xml version="1.0" encoding="utf-16"?>
-<SetupDiag xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://docs.microsoft.com/en-us/windows/deployment/upgrade/setupdiag">
+<SetupDiag xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://docs.microsoft.com/windows/deployment/upgrade/setupdiag">
```
nschonni: The `xmlns="https://docs.microsoft.com/windows/deployment/upgrade/setupdiag` doesn't make sense as it is this page, and isn't an XML schema document

PR #4102, opened 2019-06-19
chore: ms:topic -> ms.topic

nschonni:


e0i:
@Dansimp @lizap @AndreaBarr

Bringing to your attention this PR that introduces batch fixes to what appears to be punctuation errors in metadata field keys. 

Could those "errors" be intentional to avoid deployment script from parsing certain documents, maybe?

Thanks.

PR #4121, opened 2019-06-19
Change path of MDATP installation log

raeanne-marks:


e0i:
@dansimp This PR that introduces a minor fix to a path in the documentation is available for review.

PR #4191, opened 2019-06-22
fix: Replace "syntax" with "xml" code fences

nschonni:
Used find/replace regex in VS Code
Find:``` syntax(\s\n*)<\?xml
Replace:```xml$1<?xml


mypil:
@levinec - PR is ready for final review. Thanks.

DaniHalfin:
@ManikaDhiman @nickbrower - can you take a look. Since this is generated by automation, I want to make sure we're not breaking anything.

PR #4228, opened 2019-06-24
I added  asr  exclusion rules and  asr  audit mode  rules and asr dis…

RAJU2529:
01. Added Example rules for AuditMode
02. Added Example rules for Disabled
03. Added some .exe applications for exclusions and commands for these cmdlets

I dont know how to include  screenshots  in this  document

NO error1
…abled rules

Please review , after allow my additions in this document . Approve my contribution.

### DIFF
--- a/windows/security/threat-protection/windows-defender-exploit-guard/enable-attack-surface-reduction.md
+++ b/windows/security/threat-protection/windows-defender-exploit-guard/enable-attack-surface-reduction.md
@@ -146,12 +146,44 @@ Value: c:\path|e:\path|c:\Whitelisted.exe
 	```PowerShell
 	Add-MpPreference -AttackSurfaceReductionRules_Ids <rule ID> -AttackSurfaceReductionRules_Actions AuditMode
     ```
-
+	**Added Example rules for AuditMode**
+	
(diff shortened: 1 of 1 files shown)


e0i:
@levinec Would you be able to review this community contributed PR as the time permits? Thanks.

DaniHalfin:
Hi @RAJU2529 , thanks for taking the time to contribute.
Could you expand on why you've added these examples?
For all 3, did you find the initial example problematic? I'm not very sure how added multiple examples where an actual rule ID (or location) is different help clarify here.

RAJU2529:
@DaniHalfin . All examples are real, in the document. the only main command is given. But user  while reading the Document he might gets confused and he input wrong commands with applications.

Moreover, i  added almost default  ASR rules, those are already mentioned in another page.  I  collected all and added here.  After added, I conformed there is no  error messages

DaniHalfin:
Thanks for responding. 
I don't see how multiple examples of the same syntax with different rule IDs help clarify. If we get any feedback that these are confusing, I'd rather add a note pointing to the rule IDs table (that we have in another topic), and if that still isn't enough (based on feedback), we can consider adding the commands to the rule ID table. However, do note that there are other methods to add these, and we want people to have a choice rather than over-focusing on one method.
One of our voice principals is being concise, so we try to avoid redundant examples, unless there's a specific need.

e0i:
In our understanding the PR has been rejected and being closed. Thank you for taking the time to improve the Docs.

e0i:
@officedocsbot close

PR #4243, opened 2019-06-25
Updating Windows Update endpoint to list Device auth and MSA endpoint requirement

alwinv:
The Device Authentication and Microsoft Account endpoints also need to be enabled for Windows Update to function properly and find Windows Updates. This PR adds a reference to those two sections so that IT admins know to enable those endpoints as well.


e0i:
@dansimp @danihalfin Would you kindly review and merge this PR? Thanks.

PR #4291, opened 2019-06-28
add new support for SQL server 2016 sp2 and SQL 2017

intothedarkness:
this is Richard from Microsoft AppV team, we have complete the test for the latest SQL servers, and update the tech doc now.


illfated:
Good update, well done. 👍

Issue #4253, opened 2019-06-26
I want to add full commands 


RAJU2529:
I want to add the full commands with full location paths as an example.
So can i get permissions , to edit this article?.
Thanking you 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 3fc86fa9-c1cc-6b35-d3a3-ae5953953340
* Version Independent ID: 44fcb642-3ac8-cb6a-9b7c-9ce0543c1f02
* Content: [Deploy exploit protection mitigations across your organization](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-exploit-guard/import-export-exploit-protection-emet-xml)
* Content Source: [windows/security/threat-protection/windows-defender-exploit-guard/import-export-exploit-protection-emet-xml.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/windows-defender-exploit-guard/import-export-exploit-protection-emet-xml.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @levinec
* Microsoft Alias: **ellevin**

jvsam:
@officedocsbot assign @jvsam

illfated:
@RAJU2529 : Permissions to edit the public article will not be available to anyone other than Microsoft employed internal web admins belonging to the publishing team. The only thing we non-Microsoft users can do, is to open Pull Requests (referencing issue tickets or otherwise) to suggest changes to the public pages via the Github version of the page. Some of the reasons will include the necessity to verify the relevance and validity of the content, correctness and order of appearance.

RAJU2529:
@illfated. Yes sir that I knew it sir. I asked to create pull request to author

illfated:
Fair enough. That will be taken care of if you just indicate via PR title what the page is about. The Microsoft Docs team agents (@jvsam , @mypil , @e0i or new ones) will make sure that the PR gets assigned to either a) the author or b) the current owner or c) the current content department manager. If none of these are valid options, a chief tech engineer will be consulted to validate the content.

TL;DR : Post regular pull requests as before, but try to improve your PR titles (relevant description of the page and/or topic in the PR).

jvsam:
[...]

Issue #4252, opened 2019-06-26
Arranging order must be changed 


RAJU2529:
The following order makes more sense under the title "  You can enable each mitigation separately by using any of these methods"
01.Windows Security app
02.PowerShell
03.Group Policy
04.System Center Configuration Manager (SCCM)
05.Microsoft Intune
06.Mobile Device Management (MDM)

first three methods can be adopted in windows 10 pro, ent without needed windows server in the home networks.

And I also I like to add exploit protection rules created by own for the important applications. I need permission to add rules here.

Thanking you 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9e8131c6-a1f3-f711-a6b5-8f511c39c48e
* Version Independent ID: 0ba84263-f188-b0d4-c3d1-0870fcae236b
* Content: [Turn on exploit protection to help mitigate against attacks](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-exploit-guard/enable-exploit-protection)
* Content Source: [windows/security/threat-protection/windows-defender-exploit-guard/enable-exploit-protection.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/windows-defender-exploit-guard/enable-exploit-protection.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @levinec
* Microsoft Alias: **ellevin**

jvsam:
@officedocsbot assign @jvsam

jvsam:
Hi @RAJU2529 thanks for your suggestion. But can you please explain the technical rationale regarding the change in order? The documentation lists the methods and it emphasizes that they can be enabled separately, which indicates it's the user's prerogative (*User can set one or all in no particular order*). It does not state to which Windows version it's applied, so it's generally a *cater all* document. We need to understand what will be the overall impact on the quality of the documentation if the order is changed. Looking forward to hearing from you. 

Also, If you want to *suggest* exploit protection rules to be added to this document, please create a pull request (PR) and submit it for review. We always value suggestions from the community! Also, kindly review the [Microsoft Docs contributor guide](https://docs.microsoft.com/contribute/) to familiarize yourself with the process and workflow. If you already did, that's great.

RAJU2529:
@jvsam . Thanks for the reply sir. Today I create pull request .

jvsam:
[...]

PR #4294, opened 2019-06-29
added  exploit  protection rules for  important applications based on…

RAJU2529:
[protection.zip](https://github.com/MicrosoftDocs/windows-itpro-docs/files/3341926/protection.zip)
…
 request issue #4253

@jvsam. Request issue #4253. 
taken example from Windows 10 v1903 64bit buildno 18362.207.   I already added last year and keeping these settings as default.

taken much care while adding commands, so please review and test on your systems, then verify all, and approve PR .

Thanking you sir

### DIFF
--- a/windows/security/threat-protection/windows-defender-exploit-guard/enable-exploit-protection.md
+++ b/windows/security/threat-protection/windows-defender-exploit-guard/enable-exploit-protection.md
@@ -111,6 +111,335 @@ CFG will be enabled for *miles.exe*.
 
 5. Repeat this for all the apps and mitigations you want to configure. Click **Apply** when you're done setting up your configuration.
 
+###  **Enable  Migitations on  System level**
+### **The following migitations are supported on system level**
+
+**Bottomup
+  CFG
+  DEP
+  DisableNonSystemFonts
+  ForceRelocateImages
+  HighEntropy
+  SEHOP
+  TerminateOnError**
+
+  
+  To add above migitations , open powershell application with administrator  rights.
+  And type  the following commands 
+  
+  **set-ProcessMitigation -System -enable Bottomup
+    set-ProcessMitigation -System -enable CFG
+    set-ProcessMitigation -System -enable DEP
+    set-ProcessMitigation -System -enable DisableNonSystemFonts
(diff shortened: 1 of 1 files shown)


e0i:
cc @jvsam

jvsam:
Hi @levinec, does this type of contribution acceptable to this documentation https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-exploit-guard/enable-exploit-protection? I think this may not be the right place to add a long list of commands (and they need to be tested as well). Please let us know what you think so we will know how to proceed, thank you. See #4252

jvsam:
Hi @levinec and @Dansimp if time permits, can you please review this pull request and let us know if this kind of extensive content update is allowed? We appreciate your feedback, thank you.

jvsam:
Hi again @levinec and @Dansimp hoping you can provide feedback on this open pull request. Thank you.

DaniHalfin:
Sorry, but I'm not sure what issue this is addressing. The mentioned issue is basically asking to create this PR.
What are we solving with adding this information? We're adding 400+ lines with a lot of duplicate examples. 
I tend to reject this PR, because I don't understand the need for it. Please elaborate, if possible.

jvsam:
Thanks for the response @DaniHalfin. This contribution contains a list of rules put together by a community user. That's why I am asking if this kind of content is acceptable or not. If it isn't, then we can close this PR (*unfortunately I do not have that level of access*) and provide feedback.

DaniHalfin:
@jvsam - Thanks for elaborating. While I under the general intent, I'm not sure that the downside here is warranted, given that there doesn't seem to be a need for that from other members of the community.
We prefer to keep content concise when possible, and in this case this contribution would increase the size of the topic by 300%. So unless there's some additional reasoning to add this information, I tend to reject it.

jvsam:
Appreciate your input @DaniHalfin. If that's the case, will you close this pull request? Please advise. Thank you so much.

Since the community user exerted considerable effort in putting this together, it may be worth sharing the information on the TechNet forums or the TechCommunity. I'll communicate, depending on the decision with regard to this PR.

PR #4336, opened 2019-07-03
UX Name Changes

brianreidc7:
The GPO ADMX file still references the product as Windows Defender ATP. ADMX file downloaded July 3rd 2019. Therefore have updated the docs so that it refers to the correct name. When the ADMX file is updated, refresh this content.
Also updated the "Immediate Task" name to "Immediate Task (At least Windows 7)" as that is what appears in the UX.
The user account the task needs to run as is NT AUTHORITY\SYSTEM and not BUILTIN\SYSTEM (again, this is what appears when typing SYSTEM into the Task Scheduler UX. In Task Scheduler history later is shows BUILTIN\SYSTEM but that is not what the admin needs to enter to set up the scheduled task.


e0i:
@mjcaparas A new community contributed PR is available for review. Thanks.

Issue #3299, opened 2019-04-18
Interactive logon: Do not require CTRL+ALT+DEL


ronnieharper:
The text in Intune reads: Require CTRL + ALT + DEL to log on.  So confusing!

I'm going to set it to enable to test it.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 773b74b2-7dc2-d0e5-1e3b-bd6445bceda4
* Version Independent ID: 34b55057-9078-6bd0-a9ac-2f6403b84be7
* Content: [Policy CSP - LocalPoliciesSecurityOptions](https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions#feedback)
* Content Source: [windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @MariciaAlforque
* Microsoft Alias: **maricia**

illfated:
(For easier reading - direct links to the section in question)
- https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions#localpoliciessecurityoptions-interactivelogon-donotrequirectrlaltdel
- https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md#localpoliciessecurityoptions-interactivelogon-donotrequirectrlaltdel

e0i:
@officedocsbot assign @e0i

illfated:
@ronnieharper : Have you managed to test what the actual result is when you are using Intune where that text appears?

RAJU2529:
@ronnieharper. I have tested on two machines on windows 10 pro and windows 10 ent .
I enabled that particular setting,  whenever i make changes to programs. setting asked me to enter ctrl+alt+del and asked me to enter the password.
If i disabled that setting, programs does not ask  anything.
 I would recommend if you are the only user on windows 10 machine, then don't enable. Otherwise  enable it

illfated:
[...]

PR #4349, opened 2019-07-04
Update policy-csp-localpoliciessecurityoptions.md

Reeced40:
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/3299
Added NOTE.

### DIFF
--- a/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
+++ b/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
@@ -1084,6 +1084,9 @@ This security setting determines whether pressing CTRL+ALT+DEL is required befor
 
 If this policy is enabled on a computer, a user is not required to press CTRL+ALT+DEL to log on. Not having to press CTRL+ALT+DEL leaves users susceptible to attacks that attempt to intercept the users' passwords. Requiring CTRL+ALT+DEL before users log on ensures that users are communicating by means of a trusted path when entering their passwords.
 
+> [!NOTE]
+> This policy should not be enabled if there is only one user on the Windows 10 device.
+
 If this policy is disabled, any user is required to press CTRL+ALT+DEL before logging on to Windows.
 
 Default on domain-computers: Enabled: At least Windows  8/Disabled: Windows 7 or earlier.
@@ -1103,8 +1106,6 @@ Valid values:
 - 0 - disabled
 - 1 - enabled (a user is not required to press CTRL+ALT+DEL to log on)
 
-> [!NOTE]
-> I would recommend if you are the only user on windows 10 machine, then don't enable it.
 
 
 <!--/SupportedValues-->
--- a/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
+++ b/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
@@ -1107,7 +1107,6 @@ Valid values:
 - 1 - enabled (a user is not required to press CTRL+ALT+DEL to log on)
 
 
-
 <!--/SupportedValues-->
 <!--/Policy-->
 
--- a/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
+++ b/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
@@ -1085,7 +1085,7 @@ This security setting determines whether pressing CTRL+ALT+DEL is required befor
 If this policy is enabled on a computer, a user is not required to press CTRL+ALT+DEL to log on. Not having to press CTRL+ALT+DEL leaves users susceptible to attacks that attempt to intercept the users' passwords. Requiring CTRL+ALT+DEL before users log on ensures that users are communicating by means of a trusted path when entering their passwords.
 
 > [!NOTE]
-> This policy should not be enabled if there is only one user on the Windows 10 device.
+> This policy can be enabled if there is only one user using the Windows 10 device.
 
 If this policy is disabled, any user is required to press CTRL+ALT+DEL before logging on to Windows.
 
--- a/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
+++ b/windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
@@ -1103,6 +1103,10 @@ Valid values:
 - 0 - disabled
 - 1 - enabled (a user is not required to press CTRL+ALT+DEL to log on)
 
+> [!NOTE]
+> I would recommend if you are the only user on windows 10 machine, then don't enable it.
+
+
 <!--/SupportedValues-->
 <!--/Policy-->
 


e0i:
@nenonix @JohanFreelancer9 Copy editing is needed for this PR. Thanks.

illfated:
> I may be confused, but surely the policy should **only be enabled if you are the sole user** of the device? @illfated please advise

Ooh, good question. Wish I knew this topic better so I could give an affirmative answer. Though I completely agree with your suggestion to use enabled/disabled, because that is the correct binary or Boolean way to describe the options.

windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
```diff
@@ -1103,6 +1106,8 @@ Valid values:
 - 0 - disabled
 - 1 - enabled (a user is not required to press CTRL+ALT+DEL to log on)
 
+
```
illfated:
```suggestion
```

windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
```diff
@@ -1103,6 +1106,8 @@ Valid values:
...
+
```
illfated:
```suggestion
```
(redundant blank line)

windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
```diff
@@ -1084,6 +1084,9 @@ This security setting determines whether pressing CTRL+ALT+DEL is required befor
 
 If this policy is enabled on a computer, a user is not required to press CTRL+ALT+DEL to log on. Not having to press CTRL+ALT+DEL leaves users susceptible to attacks that attempt to intercept the users' passwords. Requiring CTRL+ALT+DEL before users log on ensures that users are communicating by means of a trusted path when entering their passwords.
 
+> [!NOTE]
+> This policy should not be enabled if there is only one user on the Windows 10 device.
```
illfated: > Surely the policy should **only be enabled if you are the sole user** of the device? Also since "not enabled" is the same as disabled, can we reword to use either enabled or disabled, as this is already confusing and the use of "not enabled" makes it more so @illfated please advise

Re-reading the article in question, I noticed the original text saying "If this policy is enabled on a computer, a user is not required to press CTRL+ALT+DEL to log on. Not having to press CTRL+ALT+DEL leaves users susceptible to attacks that attempt to intercept the users' passwords. Requiring CTRL+ALT+DEL before users log on ensures that users are communicating by means of a trusted path when entering their passwords.".

Based on this, I feel confident that the note could be improved as follows:

```suggestion
> This policy can be left disabled if there is only one user using the Windows 10 device.
```

nenonix: Me too! Reading the comments on the issue page made it even worse.

nenonix: @Reeced40 please review and if you agree, please apply suggestion from @illfated.

illfated: Looking back at my suggestion, I think it might be possible to improve it even more, but then to avoid the sentence being misconstrued as meaning "This policy can _only_ be enabled if there is only one user using the Windows 10 device."

nenonix: According to the issue commentary when you enable the setting then the user **is** required to press CTRL+ALT+DEL to log on, if the setting is disabled, then the requirement is disabled, the article says otherwise. Now I really am confused
The issue was really about that confusion and maybe we should clear that up first?

illfated: OK, thank you for pointing out that part too. I really need to test this on a Windows 10 Pro computer (currently at home, using my Windows 10 Home laptop).
BTW, at work I have got 2 computers in my cramped office space, 1 upgraded from Windows 7 Pro, using the Ctrl-Alt-Del method and one clean installed Windows 10 Pro where the login UI activates on click or random key press.

illfated: Oh. Please also note: back in issue ticket #3299, @ronnieharper stated:
> The text in Intune reads: Require CTRL + ALT + DEL to log on. So confusing!

Could it be that we are barking up the wrong tree here, what he actually wanted was improvement changes in Microsoft Intune or the Intune page?

nenonix: Yes, I noticed that. and then @RAJU2529 posted his test results and ...here we are

illfated: OK, fair enough. I need to test this to make sure we are "on the same page" as the actual behaviour of this policy (unless someone else have got the necessary setup to test it before me).

e0i:
@manikadhiman Please review this copy edited PR as the time permits. Thank you.

windows/client-management/mdm/policy-csp-localpoliciessecurityoptions.md
```diff
@@ -1084,6 +1084,9 @@ This security setting determines whether pressing CTRL+ALT+DEL is required befor
 
 If this policy is enabled on a computer, a user is not required to press CTRL+ALT+DEL to log on. Not having to press CTRL+ALT+DEL leaves users susceptible to attacks that attempt to intercept the users' passwords. Requiring CTRL+ALT+DEL before users log on ensures that users are communicating by means of a trusted path when entering their passwords.
 
+> [!NOTE]
+> This policy can be enabled if there is only one user using the Windows 10 device.
```
ManikaDhiman: Hi @e0i The note here contradicts with what @illfated had added earlier, which states, "This policy should not be enabled if there is only one user on the Windows 10 device." Excerpt from the customer issue: "I would recommend if you are the only user on windows 10 machine, then don't enable. Otherwise enable it," which is inline with what @illfated had added.

illfated: Sorry for my late feedback regarding this. I checked my 2 computers at work, but unfortunately I only managed to check which setting they already are using, not the behaviour of any changes to the current one. Both computers have currently got the Group Policy setting "Not configured". The odd situation is that the computer upgraded from Windows 7 Pro to Windows 10 Pro uses the Ctrl-Alt-Delete function from before the upgrade, not via Group Policy, while my main computer is using the default Windows 10 password scenario where you can mouse click or touch any keyboard key to trigger the login UI. I am guessing that this difference is due to some other UI setting from somewhere else in the OS. I think I need to find the setting which is enabled somewhere else and disable that, to see what kind of difference the Group Policy setting has, without the interference of the GUI setting.

Should I try to find out more about this on my own computers, or are you able to verify this setting without specific feedback from me?

ManikaDhiman: Hi @illfated, I have sent an email to the product team to verify this note. Meanwhile if you can verify this at your end, it would be great. 
The final version of the note, "This policy can be enabled if there is only one user using the Windows 10 device." contradicts with what was added before, "This policy should not be enabled if there is only one user on the Windows 10 device."

illfated: Thank you for your reply. I agree with your conclusion, even though it would (as you said) be great to verify the actual behaviour of this Group Policy setting when there are no other settings making changes to the UI behaviour. I will try my best to make time to test this, even if it may mean that I need to test it outside of my standard office hours. Seeing that the Ctrl-Alt-Delete behaviour can be changed in more than one place, only makes this question more intriguing to me. I will post my next feedback on this as soon as time permits (which might take more time than I would have preferred).

ManikaDhiman: Hi @illfated, @e0i, here's the response I got from product team: 

No, it’s not just for single-user systems. We explained why we dropped it from our baselines a while ago:
https://blogs.technet.microsoft.com/secguide/2015/11/18/changes-from-the-windows-8-1-baseline-to-the-windows-10-th11507-baseline/

Excerpt from the blog:
Removed settings
This section summarizes some of the 122 settings that went from a configured value to “Not Configured.” For many of these settings, such as Windows Update settings, specific configuration is best left to the organization. For others, organizations can continue enforcing settings, but we do not consider their enforcement to be necessary.
•	[…]
•	Enforcement of Ctrl+Alt+Del at logon to protect credentials from theft. This is not particularly strong protection. First, it depends on a user that’s looking at a spoofed logon screen remembering that he or she hadn’t pressed Ctrl+Alt+Del before typing a password. Second, so many apps prompt the user for the same credentials on the user’s desktop that the credentials can easily be stolen there. Third, if the adversary has gained administrative control of the computer, the “secure desktop” is no longer a protected space. Finally, with devices offering more keyboard-free logon experiences such as facial recognition, Ctrl+Alt+Del becomes an annoying interference.

So, there is no need to add any note in the docs for this policy. Hope this resolves the issue. Thanks!

illfated: Thank you for this clear feedback. I will post a comment about this and quote most (maybe all) of your comment back in the issue ticket.

e0i:
Closing the PR as no changes necessary via comments from the product team.

e0i:
@officedocsbot close

PR #4382, opened 2019-07-09
Update wcd-admxingestion.md

bhooten:
Corrected file type from ADXM to ADMX on line 54 and added preceding line break to activate markdown formatting on line 77


e0i:
@AndreaBarr Would you be able to review this PR that introduces minor typo corrections? Thanks.

PR #4419, opened 2019-07-12
fix: MD039/no-space-in-links

nschonni:
Spaces inside link text


illfated:
Nitpicking what kind of space the links can't have, just in case any MarkDown enthusiast drops by:
The square brackets around any linked text need to be next to non-space characters on both sides.

e0i:
@AndreaBarr Please review and merge this PR that introduces a minor whitespace fix in accordance with Markdown syntax. Thanks.

PR #4445, opened 2019-07-15
Add feature description and samples for AssignedAccess CSP 1809, 1903 and current Prerelease

chandde:
This is Chandler from Assigned Access feature team, updating information for GlobalProfile, AutoLaunch, FileExplorerNamespaceRestriction, which are either RTM'ed or available in pre-release (called out in the changes).


e0i:
@lomayor @dansimp

A group of authoritative changes from Assigned Access feature team is available for review. Thanks.

PR #4455, opened 2019-07-16
Remove *space from two URLs as they are wrong

brianreidc7:
The URL was * vortex-win.data.microsoft.com and * settings-win etc. But this is not valid. Also it is not *.vortex-win and *.settings-win, as the Azure Log Analytics URL page at https://docs.microsoft.com/en-us/windows/deployment/update/windows-analytics-get-started lists them like this. So this edit it to introduce consistency across articles and have correct URLs listed.


e0i:
@dansimp

The PR that introduces URL fixes is available for review. Thanks.

PR #4456, opened 2019-07-16
Clarification on path to run

brianreidc7:
Also changed the min Win10 version so its the same as the page that is on the link that follows that info.


e0i:
@Dansimp 

A new PR that introduces various clarifications is ready for review.

Thanks.

PR #4464, opened 2019-07-16
Clarifying passive mode and removing reference to LPS which is consumer

arottem:


e0i:
@dansimp The PR that introduces a clarification is ready for review. Thanks.

e0i:
@Dansimp Following up for updates.

e0i:
@Dansimp Following up for updates.

PR #4471, opened 2019-07-17
Update metadata to match correct manager of an author

e0i:
Relevant metadata changes on various articles to associate the correct manager with an author.


e0i:
@AndreaBarr @greg-lindsay Please review and merge this PR that introduces metadata updates to correctly associate an author with their manager. Thanks.

PR #4483, opened 2019-07-18
Update docs with limitation on when an option shows up as available

andreiztm:
Added note for MBEC property.


e0i:
@levinec A community contributed PR that introduces a clarifying note is available for review. Thanks.

andreiztm:
Msft here astoica@

Issue #4523, opened 2019-07-23
More information links above are not working


alnelsonau:
For example I see "[Windows Autopilot user-driven mode](user-driven.md)"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: dad887d1-fcd5-8e9d-86a2-04b9c0db692e
* Version Independent ID: aba8abe2-f7cb-cd8b-cdd4-93234396e995
* Content: [Windows Autopilot scenarios and capabilities](https://docs.microsoft.com/en-us/windows/deployment/windows-autopilot/windows-autopilot-scenarios#feedback)
* Content Source: [windows/deployment/windows-autopilot/windows-autopilot-scenarios.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/deployment/windows-autopilot/windows-autopilot-scenarios.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

mypil:
@officedocsbot assign @mypil

mypil:
@alnelsonau - Thank you for submitting feedback.

I will get this issue over to the Win10 ITPro writing team for investigation.

Thank you for reporting and making the docs better. Much appreciated.

I made a note to request the team to update this when the work is complete.

illfated:
The page https://docs.microsoft.com/en-us/windows/deployment/windows-autopilot/windows-autopilot-scenarios is now working as it should be.
The latest change to the Github version of the page was commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/edb78ff7b5c2e5be1524f3d12a2c2a63332cc602, but the problem has been mitigated directly on the docs.microsoft.com page yesterday, July 23, 2019.

mypil:
@alnelsonau - From our understanding, the issue has been resolved. If you feel it hasn't been resolved, please re-open this issue.


Thank you for your contribution to make the docs better! Much appreciated!

mypil:
@officedocsbot close

PR #4524, opened 2019-07-23
Update windows-autopilot-scenarios.md

lindspea:
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/4523
Fixed multiple links


mypil:
@greg-lindsay - PR is ready for final review. This is based on issue #4523 

Thank you.

greg-lindsay:
This particular table has already been fixed in another PR.

But, just so you know, the changes that were made here would not fix the article. The problem is that markdown links were used in an HTML table. In order for markdown links to work, the table can be converted to markdown style (this is the method that was used in the other PR). Alternatively, the links can be converted to HTML style and the HTML table style can be retained.

PR #4548, opened 2019-07-25
Update windows-10-upgrade-paths.md

andrewjohnporter:
Added information about how to to upgrade from LTSC to SAC and be able to keep applications.


illfated:
Ref. issue page #4561 (for tracking)

PR #4563, opened 2019-07-27
Update metadata to replace non-existent author

e0i:
Introduces temporary replacement for an author that could not be found within the internal directory.


e0i:
@AndreaBarr @Dansimp Please review and merge the PR containing a metadata update for non-existen author. Thanks.

e0i:
@AndreaBarr @Dansimp Following up.

PR #4597, opened 2019-08-01
Update guidelines-for-assigned-access-app.md

kellieei:
Per an ICM incident, I updated this description to be more clear about URL exceptions and adding a "." before the domain name.


e0i:
@AndreaBarr The PR that introduces a minor change to a URL quote created by a member of Microsoft Docs is available for review. Thanks.

PR #4618, opened 2019-08-03
typo

CTroessaert:
typo the Action Sataus column instead of the Action Status column


e0i:
@AndreaBarr The PR with a typo fix is available for review. Thanks.

PR #4646, opened 2019-08-07
Make the multi-user case for automatic triggers more clear

PeterSmithRedmond:
Devices with multiple always-on triggers can still only have one profile set to active out of all of the profiles for all of the users on the device.


e0i:
@dulcemontemayor @martyav @AndreaBarr  The PR that has been contributed by a MS member is available for a quick review. Thanks.

e0i:
@DulceMontemayor @martyav @AndreaBarr Following up for updates.

PR #4654, opened 2019-08-08
Update sideload-apps-in-windows-10.md

dianmsft:
Updated document to say that as of Windows Insider Build 18956, sideloading is enabled by default. Now, you can deploy a signed MSIX package onto a device without a special configuration.


MSFTTracyP:
Checking to see if this is live.

illfated:
Only on Github so far. May take up to a week to get the docs.microsoft.com page updated with this change (unless any of the central admins decides to increase the schedule frequency).

Issue #3762, opened 2019-05-21
Cloud Protection Intune Setting Contradiction


joebartlett94:
To enable cloud-delivered protection in Intune, the article here states that we should:
> On the Cloud-delivered protection switch, select Enable.

However, the [Intune documentation](https://github.com/MicrosoftDocs/IntuneDocs/blob/master/intune/device-restrictions-windows-10.md) states:

> Cloud protection: Not configured (default) allows the Microsoft Active Protection Service to receive information about malware activity from devices that you manage. Enable blocks this feature.

Please can you confirm which is correct?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5f8ccc01-1b04-e443-ef75-6ffe6c866dd5
* Version Independent ID: fa2eaa0f-db04-785a-3894-8244947b9cd5
* Content: [Enable cloud-delivered protection in Windows Defender Antivirus](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-antivirus/enable-cloud-protection-windows-defender-antivirus)
* Content Source: [windows/security/threat-protection/windows-defender-antivirus/enable-cloud-protection-windows-defender-antivirus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/windows-defender-antivirus/enable-cloud-protection-windows-defender-antivirus.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Justinha
* Microsoft Alias: **v-anbic**

e0i:
@officedocsbot assign @e0i

e0i:
Hello, @joebartlett94 

The issue is under investigation and you will be notified with any changes on the way.

Thank you for providing feedback.

Malind19:
@e0i Created PR https://github.com/MicrosoftDocs/IntuneDocs/pull/2656

e0i:
Hello @joebartlett94 

Based on our investigation, the Intune documentation is correct. Find the quote below,

> The settings documented are correct. This Cloud Protection setting uses the Defender/AllowCloudProtection CSP. The CSP description states:

> 0 – Not allowed.
> 1 (default) – Allowed.

> In Intune, Not configured is the default, which means this setting is allowed. The use of the word "Enable" in the UI definitely makes it confusing.

Please let us know whether this information clarifies your query. Thanks.

[...]

PR #4694, opened 2019-08-14
Fix Intune "Cloud-delivered protection" setting

joebartlett94:
Changed guidance on Intune's "Cloud-delivered protection" toggle to "Not configured" as per the [Intune documentation](https://github.com/MicrosoftDocs/IntuneDocs/blob/master/intune/device-restrictions-windows-10.md)/issue #3762 .


e0i:
@dansimp The community contributed PR that addresses issue #3762 is available for review. The PR contents are based on findings from a previous investigation quoted below,

>   The settings documented are correct. This Cloud Protection setting uses the Defender/AllowCloudProtection CSP. The CSP description states:

>   0 – Not allowed.
>   1 (default) – Allowed.

>   In Intune, Not configured is the default, which means this setting is allowed. The use of the word "Enable" in the UI definitely makes it confusing.

e0i:
@Dansimp Following up for updates.

PR #4743, opened 2019-08-20
Replace a retired author handle from metadata

e0i:
Introduces temporary replacement for an author, _"Mir0sh"_, that could not be found within the internal directory.


e0i:
@Dansimp Please review and merge the PR containing a metadata update for a non-existent author. Thanks.

PR #4795, opened 2019-08-27
Fixed types for the parameters

guregini:
Some of the types were written in the name column


e0i:
@mjcaparas The PR fixes the types of parameters for the request parameters in the page, _" Pull Microsoft Defender ATP alerts using SIEM REST API"_ is available for review. Thanks

PR #4804, opened 2019-08-28
CI 105177

Deland-Han:
New topics about troubleshooting Windows startup issues.


illfated:
Regarding the name and title of the file

## Introduction of page file

Even though English is not my first language, according to my experience the expression "introduction of" means "first appearance of" or "first occurrence of", similar to "the feature was first introduced [ ... ] (time or place of introduction)".

In my opinion (again, from my experience), it reads better in English if it is written like this (or similar to this):

## An introduction to the page file

Feel free to ignore or debate my suggestion. I just wanted to express my reaction to how your title reads for some of us.

Deland-Han:
@illfated I agree with you. It could be better to use "to" instead of "of" here. Meanwhile, I think we can remove the article "an" from the beginning of titles. Therefore, I'd change it to "Introduction to the page file". Looks okay to you?

illfated:
Sure, fine with me, even though I view it as only one out of many different introduction pages of various quality, type or content, around the internet.
Of course, in this context, here in this page context, it is OK to leave the title open for interpretation, because the chance that someone would look around for other web pages concerning page files is very low.
From a prosaic writer's point of view, it still feels more natural to say "An introduction to the page file" -- simply because we know that the subject has definitely been described elsewhere before.

e0i:
@Dansimp The PR that introduces various new articles written by an internal MS contributor is available for review.

Deland-Han:
Thank you. @illfated and @e0i. I have requested an editor to help proofread and edit these articles, and then re-pushed the edited version.

@Dansimp  could you help review the articles and merge this pull request if they look good to you?

illfated:
Good job. Thanks for correcting the file name too, in addition to the link text (as discussed a week ago).
I thought about mentioning the file name a week ago, but something else caught my attention.

PR #4816, opened 2019-08-29
Clarify the registry key needed to set tags

brianreidc7:


e0i:
@mjcaparas The PR that clarifies the registry keys within the documentation is available for review. Thanks.

e0i:
@mjcaparas Following up for updates. Thanks.

PR #4947, opened 2019-09-16
SMB 3.x support AES-128 encryption

nedpyle:
SMB 3.x is the remote file system protocol of Windows, and has supported AES-128 encryption via BCRYPT since Windows 8/WS2012.


e0i:
@dulcemontemayor The PR that has been provided by an internal MS contributor is available for review. Thanks.

e0i:
@DulceMontemayor Following up for updates.

PR #4970, opened 2019-09-19
Update microsoft-defender-advanced-threat-protection.md

ErikMoreau:
replaced link to ATP trial (ended up on Windows 10 devices for business page with no link to ATP Trial

### DIFF
--- a/windows/security/threat-protection/microsoft-defender-atp/microsoft-defender-advanced-threat-protection.md
+++ b/windows/security/threat-protection/microsoft-defender-atp/microsoft-defender-advanced-threat-protection.md
@@ -19,7 +19,7 @@ ms.topic: conceptual
 
 # Microsoft Defender Advanced Threat Protection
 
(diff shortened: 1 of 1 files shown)


illfated:
Out of curiosity, would it be useful in any way to skip the `/en-us/` part of the link? I mean, I tested removing it and got mostly the same link in return (https://www.microsoft.com/en-us/microsoft-365/windows/microsoft-defender-atp?rtc=1). I am guessing there are no translated pages for Microsoft ATP trial yet, but would it be practical at a later date, when the page gets translated?

e0i:
@mjcaparas The PR fixes a link to ATP trial is available for review. Thanks.

e0i:
@mjcaparas Following up for updates.

Dansimp:
We did a global search & replace to fix this URL as it shows up often in our docs. All instances should be fixed now, this PR is no longer necessary.

illfated:
The docs.microsoft.com pages are updated with the correct link. I find it interesting to see that the `public` branch is updated, but not synced to the `master` branch yet. Could be a useful observation with regards to which branch to open future PRs to.

Issue #4954, opened 2019-09-17
Support for Studio 2? 


luriep:
The release notes say that in Version 3.2.75.0, support for Surface Studio 2 was added.  However, in the main list of Compatible Surface devices, the Studio 2 is not listed.  Please add it to the list.  

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 93a0e269-be9b-781a-b9b4-52b94e68618b
* Version Independent ID: 725c2d4a-5397-6c67-7605-93a8cd40cf08
* Content: [Microsoft Surface Data Eraser (Surface)](https://docs.microsoft.com/en-us/surface/microsoft-surface-data-eraser#feedback)
* Content Source: [devices/surface/microsoft-surface-data-eraser.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/devices/surface/microsoft-surface-data-eraser.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

mypil:
@luriep  - Thank you for submitting feedback.

I will get this issue over to the Win10 ITPro writing team for investigation.

Thank you for reporting and making the docs better. Much appreciated.

I made a note to request the team to update this when the work is complete.

mypil:
@luriep - From our understanding, the issue has been resolved but it may take a few days for the merged content to appear in the article. If you feel it hasn't been resolved, please re-open this issue.

Thank you for your contribution to make the docs better! Much appreciated!

PR #4997, opened 2019-09-23
added support of Surface Studio 2 to the list of compatible devices

MaratMussabekov:
#4954


mypil:
@Dansimp - PR is ready for final review based on issue #4954. 

Thank you.

cc: @kenwith

PR #5009, opened 2019-09-23
Schedule task hidden from standard users

ShrCaJesmo:
Adds note that the scheduled task that triggers GPO enrollment isn't visible to standard users so to confirm the task was created, admin creds should be used


ShrCaJesmo:
I may have gotten this wrong - I need to test a bit further. Either the task didn't appear because it's not visible to Standard users, or because it only creates the task in the session signed in when the gpo applies. If someone could validate either way that'd be great but otherwise I'll update this once I confirmed the outcome I saw wasn't caused by another mechanism

windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md
```diff
@@ -198,6 +198,10 @@ To collect Event Viewer logs:
     The auto-enrollment process is triggered by a task (Microsoft > Windows > EnterpriseMgmt) within the task-scheduler. This task appears if the *Enable automatic MDM enrollment using default Azure AD credentials* group policy (Computer Configuration > Policies > Administrative Templates > Windows Components > MDM) is successfully deployed to the target machine as shown in the following screenshot:
     ![Task scheduler](images/auto-enrollment-task-scheduler.png)
 
+
```
illfated: cosmetic/codestyle: This extra blank line is redundant, it does not increase blank space after the image.
```suggestion
```

illfated:
The note annotation looks good so far, although I do not know the topic in question well enough to comment on the content.

ShrCaJesmo:
Thank you! I'll test a few more variants to make sure I've got it right then update here with my results to validate

illfated:
BTW, is this PR a response to any existing open issue ticket for this page?

ShrCaJesmo:
It's not, no - should I have made an issue first?

illfated:
No, it is just fine as it is. It would simply be useful to refer to any existing issue tickets if they had been present already.

ShrCaJesmo:
So part of the reason I hesitated was that I noticed it had created a new folder with the standard user's SID at the root of Task Scheduler that contained an EnterpriseManagement folder, so I suspected that account could see the task there. Nope. It's still invisible. Re-ran gpupdate from standard user, closed and reopened task scheduler, the EnterpriseManagment folders both look empty as a standard user. I can see other tasks just fine as standard user, so it's not that Task Scheduler isn't working right, it looks like from a standard user perspective the tasks simply aren't available. Revalidated, admin can see them just fine. It seems it really is just hidden from standard users.

illfated:
Glad to hear it. Looks like solid research to me. I will go ahead and approve this PR (even though I'm not part of the MS Docs team), just to note that it is ready for final review.

illfated:
@e0i : Will you notice the author(s) and manager of this document that it is ready for final review?

e0i:
@manikadhiman Would you be able to review this PR that provides a clarifying note regarding scheduled tasks? Thanks.

e0i:
@ManikaDhiman Following up for updates.

Issue #4995, opened 2019-09-23
JA machine translation of numbers is incorrect


takondo:
The machine translated version's Logon Type table is terrible. It doesn't seem to recognize the numbers, as the text for Logon Type is jibberish that makes no sense. 

---
#### ドキュメントの詳細

⚠ *このセクションを編集しないでください。 docs.microsoft.com で必須です ➟ GitHub の問題のリンク。*

* ID: aefbbdcc-606c-276c-47ee-fe2b2cb65c61
* Version Independent ID: e0979867-e6bc-7fb9-1279-646b319823b0
* Content: [4624アカウントが正常にログオンしました。 (Windows 10)](https://docs.microsoft.com/ja-jp/windows/security/threat-protection/auditing/event-4624)
* Content Source: [windows/security/threat-protection/auditing/event-4624.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/auditing/event-4624.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

illfated:
Looks like the English version of that table can be improved with code block fences (or simply MD code markers (back ticks)) to keep the numbers from being translated.

At the same time, should column 2 (**Logon Title**) also retain the original English names as code, not to be translated?
***
| Logon Type | Logon Title       | Description                                                                                                                                                                                                                                                                                                                |
|:----------:|-------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|  `2`       | Interactive       | A user logged on to this computer.                                                                                                                                                                                                                                                                                         |
|  `3`       | Network           | A user or computer logged on to this computer from the network.                                                                                                                                                                                                                                                            |
|  `4`       | Batch             | Batch logon type is used by batch servers, where processes may be executing on behalf of a user without their direct intervention.                                                                                                                                                                                         |
|  `5`       | Service           | A service was started by the Service Control Manager.                                                                                                                                                                                                                                                                      |
|  `7`       | Unlock            | This workstation was unlocked.                                                                                                                                                                                                                                                                                             |
[...]

PR #5026, opened 2019-09-24
Auditing: add MD code block to table keywords

illfated:
**Description:**

This PR aims to block keywords and values from being translated to another language, keeping the values and keywords intact, in response to the windows-itpro-docs issue ticket #4995.

**Proposed changes:**

- Surround Logon Type values and Logon Title keywords with MD code block markers (back ticks) to keep them from being translated by MT.
- Add XML indentation to the XML for the layout to be shown properly.
- Center the numerical values in column 1 (Logon Type).

Thanks to @takondo for pointing out the uselessness in translating these parts of the table when the document is machine translated.

**issue ticket reference or closure:**
Ref. issue ticket #4995

(The ticket can be closed when this commit is successfully migrated and shown to be effective against machine translation.)

### DIFF
--- a/windows/security/threat-protection/auditing/event-4624.md
+++ b/windows/security/threat-protection/auditing/event-4624.md
@@ -33,55 +33,55 @@ This event generates when a logon session is created (on destination machine). I
 <br clear="all">
 
 ***Event XML:***
-```
-- <Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
-- <System>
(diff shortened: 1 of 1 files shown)


illfated:
@kenwith : 👍 Nice, thank you for your quick response and merge. All we need to do now is to wait and see if this change manages to achieve the expected results after migration to docs.microsoft.com and its translated pages.

PR #5028, opened 2019-09-24
Update security-compliance-toolkit-10.md

jmunck:
Removed unsupported baseline and downloads for Windows 10 1511 and Windows 10 1703.  Removed Office 2016 baseline and replaced with Office365 ProPlus (Sept 2019)


e0i:
@dulcemontemayor The PR that has been introduced by an internal MS contributor is available for review. Thanks.

PR #5029, opened 2019-09-24
Removed "About" Tab from the UEFI menu

actena-msft:
- Removed "About" Tab and section
- Added FW update bar colors
- Renamed files to follow number sequence.


e0i:
@dansimp The PR that has been contributed by an internal MS member is available for review. Thanks.

PR #5051, opened 2019-09-27
update an old article support links.

Deland-Han:
based on the request from Scott McArthur (NC)


e0i:
@dansimp The PR that has been provided by an internal MS contributor is available for review. Thanks.

trinhtuanvy1988:
> based on the request from Scott McArthur (NC)

please help setup

PR #5082, opened 2019-10-01
Update volume-activation-windows-10.md

baardhermansen:
Updated wording.
Updated links. 
A bit of reformatting.

### DIFF
--- a/windows/deployment/volume-activation/volume-activation-windows-10.md
+++ b/windows/deployment/volume-activation/volume-activation-windows-10.md
@@ -1,5 +1,5 @@
 ---
-title: Volume Activation for Windows 10 (Windows 10)
+title: Volume Activation for Windows 10
 description: This guide is designed to help organizations that are planning to use volume activation to deploy and activate Windows 10, including organizations that have used volume activation for earlier versions of Windows.
 ms.assetid: 6e8cffae-7322-4fd3-882a-cde68187aef2
 ms.reviewer: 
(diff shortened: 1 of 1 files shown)


illfated:
Rich Text Diff view, for easier overview of the changes:
https://github.com/MicrosoftDocs/windows-itpro-docs/pull/5082/files?short_path=f05069d#diff-f05069d26735f8a569c41af2abfa69e4

@baardhermansen : Why replace the `em` style dash (old line number 46) with a regular dash (new line number 48)?
From a reader's point of view, I don't see the improvement, it looks more like an unnecessary simplification.

At first glance, the rest of the content looks fine, but I will still recommend a copy review.

baardhermansen:
@illfated The replacement of the dash is because the previous version look like a hyphen, thus making reading slightly more cumbersome.

illfated:
Oh. Very well, fair enough. I looked up "hyphen or em dash" just now and it appears that _hyphen_ is the term I meant instead of "dash" as in regular usage of a hyphen. I presume my preference comes from seeing the em dash being used in formal writing, whereas the hyphen is used in modern writing more like an alternative type of spacing.

e0i:
@greg-lindsay The PR provides minor Markdown formatting fixes. Please review and merge as the time permits. Thanks.

baardhermansen:
I'm not really a fan of the hyphen any way, would actually have preferred using commas, but YMMV.

illfated:
Fair enough. As a matter of fact, the same part of the sentence does in fact work completely without hyphens, commas or em dash if that looks more readable:

> Volume activation and the need for activation itself is not new, and this guide does not review all of its concepts and history.

The importance of separation or differentiation of the inserted clause is usually the author's privilege, depending on the intended focus and overall meaning.

e0i:
@greg-lindsay Following up for updates. Thanks.

PR #5095, opened 2019-10-02
Update windows-security-baselines.md

jmunck:
Updated the applies to section based on currently supported baselines


e0i:
@DulceMontemayor Following up for updates.
cc @Dansimp

PR #5174, opened 2019-10-14
Typo "**user@example.com**"→"**user\@example.com**"

hyoshioka0128:
bold with escape characters

https://docs.microsoft.com/ja-jp/surface-hub/surface-hub-2s-setup


illfated:
This update is actually needed both for the en-us page and all the other language pages.
See step 3 in the first section for any language version of the page:
- https://docs.microsoft.com/en-us/surface-hub/surface-hub-2s-setup
- https://docs.microsoft.com/ja-jp/surface-hub/surface-hub-2s-setup

CC: @kenwith , @e0i

Dansimp:
we will fix the en-us version of the topic first. We have automation in place to pick up changed topics and localize them. So, the ja-jp versions should get updated a few days after we update the en-us version of the topic.

PR #5197, opened 2019-10-16
Adding `l` in Application

muffins:


windows/security/threat-protection/windows-defender-application-control/audit-windows-defender-application-control-policies.md
```diff
@@ -21,7 +21,7 @@ ms.date: 05/03/2018
...
+Running Application Control in audit mode allows administrators to discover any applications that were missed during an initial policy scan and to identify any new applications that have been installed and run since the original policy was created. While a WDAC policy is running in audit mode, any binary that runs and would have been denied had the policy been enforced is logged in the **Applications and Services Logs\\Microsoft\\Windows\\CodeIntegrity\\Operational** event log. When these logged binaries have been validated, they can easily be added to a new WDAC policy. When the new exception policy is created, you can merge it with your existing WDAC policies.
```
illfated: Good find, well spotted. May I suggest adding **bold text** for the feature name, just to separate it better from the first word of the sentence?
```suggestion
Running **Application Control** in audit mode allows administrators to discover any applications that were missed during an initial policy scan and to identify any new applications that have been installed and run since the original policy was created. While a WDAC policy is running in audit mode, any binary that runs and would have been denied had the policy been enforced is logged in the **Applications and Services Logs\\Microsoft\\Windows\\CodeIntegrity\\Operational** event log. When these logged binaries have been validated, they can easily be added to a new WDAC policy. When the new exception policy is created, you can merge it with your existing WDAC policies.
```

Issue #4261, opened 2019-06-26
Clarification on wildcard


nluck:
The section on wildcards lists few examples: It does look like the wildcard syntax follows the AppLocker path variable syntax.
Are other AppLocker path variables (such as %PROGRAMFILES% and %HOT%) also implemented and do they have the same effect?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 94cfbb2d-7562-dd7a-ce15-a3e30ad88544
* Version Independent ID: 535b2189-c7c3-ae3c-b939-62bfdaf8d28d
* Content: [Windows Defender Application Control path-based rules (Windows 10)](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-control/create-path-based-rules#feedback)
* Content Source: [windows/security/threat-protection/windows-defender-application-control/create-path-based-rules.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/windows-defender-application-control/create-path-based-rules.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @mdsakibMSFT

jvsam:
@officedocsbot assign @jvsam

jvsam:
Hi @nluck I believe your question is somewhat answered by this note in this [doc](https://docs.microsoft.com/windows/security/threat-protection/windows-defender-application-control/create-path-based-rules#feedback) where it says "*AppLocker uses custom path variables for well-known paths, such as Program Files and Windows.*" And since WDAC now supports path-based rules beginning from 1903, then both now support *some* wildcards. If you wish to confirm if the support extends to some variables that are not in this documentation, I would recommend contacting [Windows 10 Support](https://support.microsoft.com/hub/4338813/windows-help?os=windows-10) to get an immediate answer. Apologies, but please be informed that this site is intended to help improve the quality of the technical documentation of Microsoft Docs through user contributions and feedback. If you have suggestions related to this documentation, please let us know here and we will submit them for review. Same goes if you get feedback from Support, please do keep us posted here on the resolution or any relevant information that can be called out in the docs, *for example, additional path rules that are currently not available in this documentation*. Our main goal is to ensure users have the best Windows 10 experience by providing useful, accurate and up-to-date documentation. 

Thanks for being part of the Microsoft Docs community!

nluck:
Thank you for your reply. 

[...]

PR #5200, opened 2019-10-16
Update filepath rules documentation

brbrahm:
Clarify wildcard syntax and new -Level options


brbrahm:
I'm a WDAC feature team member. These updates are in order to resolve issue #4261

illfated:
The only constructive detail I could imagine would be nice (but maybe cosmetic only?) would be to have the cardinal numbers 0-19 in **Table 2. Windows Defender Application Control policy - policy rule options** changed to ordinals (0.-19.) for better visual spacing. On the other hand, maybe that could change the logical interpretation of the numbers, I am not sure. What do you think?

( I presume the choice of starting the numbering from 0 signifies that the first one (0) is some kind of major default rule, otherwise I would have questioned why not use 1-20.)

brbrahm:
Hi @illfated -- I agree that changing to ordinals might look better, but might also cause confusion. In order to set Enabled:UMCI, you'd need to run the command Set-RuleOption 0. Our current documentation also matches the style of the -help output users can see in PowerShell.

illfated:
Very well, fair enough. Thank you. Based on your feedback, I see no need to pursue my previous train of thoughts.

@e0i : May we have a final review & merge by the document author or manager?

e0i:
cc @jvsam

jvsam:
Hi @Dansimp this pull request is ready for your review and approval. Please check, thanks.

brbrahm:
Hi @Dansimp-- friendly request to review this PR

PR #5218, opened 2019-10-18
DnsAdmins and DnsUpdateProxy are not well-known

piaudonn:
I added the following: "This is not a well-known SID. The RID part may vary from domain to domain." for both of them and replace the well-known SID by "It may vary from environment to environment" because the RID selection depends on other factors (when the DNS service is installed for example).

### DIFF
--- a/windows/security/identity-protection/access-control/active-directory-security-groups.md
+++ b/windows/security/identity-protection/access-control/active-directory-security-groups.md
@@ -1345,7 +1345,7 @@ This security group has not changed since Windows Server 2008.
 
 Members of the DnsUpdateProxy group are DNS clients. They are permitted to perform dynamic updates on behalf of other clients (such as DHCP servers). A DNS server can develop stale resource records when a DHCP server is configured to dynamically register host (A) and pointer (PTR) resource records on behalf of DHCP clients by using dynamic update. Adding clients to this security group mitigates this scenario.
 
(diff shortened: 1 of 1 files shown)


illfated:
Rich Text Diff view: https://github.com/MicrosoftDocs/windows-itpro-docs/pull/5218/files?short_path=3046dde#diff-3046dde9adc139f8511742d52a29954b

I am looking forward to feedback from the document author, because to me it looks like the text "This is not a well-known SID. The RID part may vary from domain to domain." seems to contradict most of the Attribute text "Well-Known SID/RID", making the whole table look redundant. (I could be wrong.) Wouldn't it be just as useful to add the text "The RID part may vary from domain to domain." as an extra explanatory text, or maybe replace the RID part with something like `<variable RID>` and possibly change the attribute text to only `Well-Known SID` ?

If my take on this modification is too far off to be appropriate, please forgive my lack of knowledge on this topic.

piaudonn:
I like your suggestion 👍  <variable RID>

illfated:
Thank you for the nice reply. Depending on feedback from the document author (@DulceMontemayor ), you may choose between editing the PR content yourself or a suggestion comment from me directly in the PR content, in case that is an easier option for you to implement. (I will wait for more feedback in the meanwhile.)

piaudonn:
Note that I believe that's the same with WinRMRemoteWMIUsers_. The RID is not necessary 1000.

illfated:
Sounds fair, so if you can document it or refer to material where that has been tested and confirmed, you may want to propose a similar change there too.

e0i:
@Dansimp The PR that has been created by a MS contributor that has passed copy review is ready for final review & merge. Thank you.

PR #5220, opened 2019-10-18
Update white-glove.md

doctorscripto:
From a discussion with a customer who is Implementing this technology, they have confirmed it did not work in a Hybrid Configuration until Writeback for Devices was enabled in Azure AD Connect


greg-lindsay:
I'm checking with the product team before OK'ing this change.

greg-lindsay:
Hi, according to the product team this setting is not required by Autopilot. It isn't clear why this helped in the current scenario. Further analysis is necessary - but we can't add this as a requirement.

PR #5272, opened 2019-10-25
Removed unneeded character

markaldridge:
Removed a single character that wasn't meant to be there.

### DIFF
--- a/windows/deployment/upgrade/windows-error-reporting.md
+++ b/windows/deployment/upgrade/windows-error-reporting.md
@@ -20,7 +20,7 @@ ms.topic: article
 **Applies to**
 -   Windows 10
 
->[!NOTE]
+> [!NOTE]
 > This is a 300 level topic (moderately advanced).  
 > See [Resolve Windows 10 upgrade errors](resolve-windows-10-upgrade-errors.md) for a full list of topics in this article.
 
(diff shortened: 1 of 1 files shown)


windows/deployment/upgrade/windows-error-reporting.md
```diff
@@ -29,7 +30,7 @@ When Windows Setup fails, the result and extend code are recorded as an informat
 To use Windows PowerShell, type the following commands from an elevated Windows PowerShell prompt:
 
 >[!IMPORTANT]
->}The following source will be available only if you have updated from a previous version of Windows 10 to a new version. If you installed the current version and have not updated, the source named **WinSetupDiag02** will be unavailable.
+>The following source will be available only if you have updated from a previous version of Windows 10 to a new version. If you installed the current version and have not updated, the source named **WinSetupDiag02** will be unavailable.
```
illfated: Good catch. Since we are already here making an edit, could we also add a MarkDown compatibility space?
```suggestion
> The following source will be available only if you have updated from a previous version of Windows 10 to a new version. If you installed the current version and have not updated, the source named **WinSetupDiag02** will be unavailable.
```
(if you prefer to edit the file directly, please add the same type of space in the previous line.)

markaldridge: I've added these now, thanks for the suggestion.

PR #5310, opened 2019-10-29
Update add-devices.md

andresravinet:
Windows Autopilot information page URL is out of date.


greg-lindsay:
Thanks for pointing out that the link is out of date, however I can't approve this change as is. The text refers to a section that isn't present on the new link.  We need to find a link that displays the list of supporting OEMs.  I'm looking for this now.

Issue #4973, opened 2019-09-19
Table 7. User’s or Computer’s account UAC flags. - where is it?


MrKragh:
In the article there is referred to "Table 7. User’s or Computer’s account UAC flags", but nowhere to find. It's needed to translate the events and alert based on content.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ccd81722-4e7e-1da0-e3f0-73fafd0c0ac5
* Version Independent ID: 7899a9fa-e5b0-9563-de93-b3a40b3494ff
* Content: [4738(S) A user account was changed. (Windows 10)](https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4738)
* Content Source: [windows/security/threat-protection/auditing/event-4738.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/auditing/event-4738.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

illfated:
This is the referenced table, although any links to it is missing in the Auditing Event document pages:
- https://support.microsoft.com/en-us/help/305144/how-to-use-useraccountcontrol-to-manipulate-user-account-properties

Web search will also give you some non-Microsoft pages containing the same table & information.
Search term: `table 7 UserAccountControl Attribute/Flag Values`

MrKragh:
I have no trouble finding the official UAC list.
Problem is under the field "User Account Control" they reffer to this table.
Values in this field are like the following "%2048", and there is no translation table, other than what you can find from different searches which isn't an full list.

illfated:
Fair point. I admit that it looks like the text is incomplete or that some links to actual reference tables are missing throughout.

j0rt3g4:
Solving Table 7 references on PR #5315

e0i:
@MrKragh 

With the help of your feedback, the content has been updated with relevant changes accordingly. Thanks.

PR #5314, opened 2019-10-30
Issue#4973

j0rt3g4:
Adding Link to the #4973 
showing where the whole table is and also it's explained with an example so it's a good reference.


j0rt3g4:
Closed, it will be substituted for 1 PR that brings a single file change.

PR #5330, opened 2019-11-01
Updating benign smart card errors

mapalko:
Updating error 610 and 621 as benign since they do not impact end use of smart cards in most scenarios.


mapalko:
@Dani

Issue #5206, opened 2019-10-17
Ports required to access these url


mzafzal1:
It is not mentioned anywhere which ports need to be opened? http(80), https(433) or both? 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f86b0db3-f979-bd1f-78a4-41247510721a
* Version Independent ID: 5289bb5f-f015-7f38-05a1-26a615ae07aa
* Content: [Configure and validate Windows Defender Antivirus network connections](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-antivirus/configure-network-connections-windows-defender-antivirus)
* Content Source: [windows/security/threat-protection/windows-defender-antivirus/configure-network-connections-windows-defender-antivirus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/windows-defender-antivirus/configure-network-connections-windows-defender-antivirus.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

e0i:
@mzafzal1 

The issue has been noted and the document will be revised in the light of your feedback.

Thanks.

joinimran:
@mzafzal1  - All communications are done using secure channels and you need to unblock port 443 for these URLs.
Adding the same info in the doc as well.

Thanks

e0i:
@mzafzal1 

With the help of your feedback, the content has been updated with relevant changes accordingly. Thanks.

PR #5339, opened 2019-11-01
Added a port

joinimran:
Added port number which these URLS are using.

Problem: https://github.com/MicrosoftDocs/windows-itpro-docs/issues/5206


illfated:
Note to myself: The added sentence could do with an English fluency (grammar improvement) update.

PR #5363, opened 2019-11-06
Minor language change in second sentence

bmkaiser:
Original : when it is not IS use
Proposed : when it is not IN use


illfated:
@bmkaiser : Good find, thank you for the typo correction.
Slightly funny how that typo turned into a grammar error.

Issue #5368, opened 2019-11-06
Command shown is incorrect


nzrb:
The command to enable on a single machine is wrong, it should be

`mdatp --edr --earlyPreview true`

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c91a5cfb-d32d-3864-c1d5-aec39558f66f
* Version Independent ID: dbd2647e-88a6-ff22-4eab-ef628ac2e605
* Content: [Microsoft Defender ATP for Mac](https://docs.microsoft.com/en-nz/windows/security/threat-protection/microsoft-defender-atp/endpoint-detection-response-mac-preview#feedback)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/endpoint-detection-response-mac-preview.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/endpoint-detection-response-mac-preview.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

RyanMorash:
I've submitted pull request #5378 to resolve this.

illfated:
The PR has been rejected. Please see the closing comment in PR #5378 .

e0i:
Closing as the referenced PR has been rejected by the document author:

**levinec** wrote:
> According to the PM, this change is not necessary. "There is no need. We added backward supportability so that both modes work. Changing doc at this point would probably cause more confusion since not all customers have the same version."

_(via https://github.com/MicrosoftDocs/windows-itpro-docs/pull/5378#issuecomment-552580165)_

codeskipper:
Not sure what you meant with backwards compatibility, maybe you are referring to a version not yet released?

When following the instructions above, I verified I'm running  version 100.72.15 as it says, running the command as it says in the instructions 
> mdatp --edr --early-preview true
results in 
Failed: unrecognised option '--early-preview'

When running 
mdatp --edr --earlyPreview true
it asks for my admin credentials, and then the change to preview mode is made as expected

As it stands, the instructions do not apply to the available version

shawnharry:
Codes Skipper is correct mdatp --edr --early-preview true does not work on High Sierra 10.13.6. mdatp --edr --earlyPreview true does work correctly though.

illfated:
[...]

PR #5378, opened 2019-11-07
Corrected the command line flag

RyanMorash:
Corrected the command line flag to manually enable the EDR preview.


RyanMorash:
This resolves issue #5368

levinec:
According to the PM, this change is not necessary. "There is no need. We added backward supportability so that both modes work. Changing doc at this point would probably cause more confusion since not all customers have the same version."

illfated:
Ref. #5507

PR #5395, opened 2019-11-08
update TOC to match article title

RowdyChildren:
updated TOC to switch Network security: Configure encryption types allowed for Kerberos to the correct article title, previously it said Win7 Only at the end


illfated:
Ref. related PR #5396

PR #5396, opened 2019-11-08
Updated title of article to match actual title

RowdyChildren:
Updated title in table to match actual article title for the Network security Configure encryption types allowed for Kerberos article


illfated:
Note: This is a metadata update, which is not the visible title on the docs page as such, but still a good update for internal usage.

PR #5483, opened 2019-11-20
Update HostedInstall for EnterpriseModernAppManagement CSP

dianmsft:
The value for deployment options are not listed.  There are also other parameters for install not listed for the HostedInstall command.


Dansimp:
Do not merge until this is resolved: - ValidateDependencies="1" - (need to check with dev on the actual functionality)

dianmsft:
Updated to add the proper description.

illfated:
Odd behaviour in the Rich Text Diff view (https://github.com/MicrosoftDocs/windows-itpro-docs/pull/5483/files?short_path=8774026#diff-877402675b48cb174de44da8cec1e748) :
the PR content seems to skip all natural line breaks (ordinary NewLine), making the added text section a continuous text section. I do not know if this is the actual behaviour when migrated to docs.microsoft.com , but it looks quite impractical at the moment (if the Rich Text Diff view can be trusted).

illfated:
Very good, thanks for inserting the bullet point list. 
That resolved the issue of contiguous text without line breaks.

PR #5499, opened 2019-11-21
Remove the duplicate 'will'

kenjohnson03:


illfated:
`<irony>` I wonder if "We will, we will Rock You" was playing in the background when that line was written. 😄 `</irony>`

PR #5537, opened 2019-11-27
add information about smartscren log enable method

MS-binche:
Windows event log for SmartScreen is disabled by default, users can use Event Viewer UI to enable the log or use the command line to enable it:

```
wevtutil sl Microsoft-Windows-SmartScreen/Debug /e:true
```


MS-binche:
@Dansimp @DaniHalfin @Justinha 
Please approve and merge.

PR #5555, opened 2019-11-29
Typo fix: "better then" -> "better than"

HumanEquivalentUnit:


illfated:
Thank you. I really approve of these grammar corrections. 👍
I see too many misspellings online as it is already, so every little correction helps.

Issue #5562, opened 2019-11-29
Erreur 404


Efflamm:
Lien invalide :
https://www.kingston.com/us/usb/bootable/ikw700

---
#### Détails du document

⚠ *Ne pas modifier cette section. C’est obligatoire pour docs.microsoft.com ➟ Liaison des problèmes GitHub.*

* ID: 93868b54-1342-6a97-ea84-ab75066c837b
* Version Independent ID: 8ee0f61c-67cd-9df6-a8e9-b854d8134587
* Content: [Présentation de la fonctionnalité Windows to Go (Windows 10)](https://docs.microsoft.com/fr-fr/windows/deployment/planning/windows-to-go-overview#feedback)
* Content Source: [windows/deployment/planning/windows-to-go-overview.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/deployment/planning/windows-to-go-overview.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

jvsam:
@officedocsbot assign @jvsam

RAJU2529:
@jvsam. afternoon I will create PR

illfated:
@Efflamm : Où avez-vous trouvé ce lien invalide?
Dans quelle section ou paragraphe se trouve-t-il?

illfated:
Aha, je vois. La section IronKey entière souffre de problèmes de format de liaison, mais l'erreur 404 réelle est la première ligne de lien. Voir l'exemple de texte copié ci-dessous:

>  /!\ Avertissement
>
> L’utilisation d’un lecteur USB qui n’a pas été certifié n’est pas prise en charge.

> **Espace de travail IronKeyhttp://www.ironkey.com/windows-to-go-drives/ironkey-workspace-w700.htmlW700 ()** ⬅️ 
> Espace de travail IronKeyhttp://www.ironkey.com/windows-to-go-drives/ironkey-workspace-w500.htmlW500 ()
> Espace de travail IronKeyhttp://www.ironkey.com/windows-to-go-drives/ironkey-workspace-w300.htmlW300 ()
> Espace de travail DataTraveler Kingston pour Windows tohttp://www.kingston.com/wtg/Go ()

illfated:
The en-us page looks something like the following:
(Please note, I have added the fwlinks in parentheses at the end of the rows.)
***
/!\ Warning

[...]

PR #5566, opened 2019-11-30
old link is replaced  by new link

RAJU2529:
as per the user report #5562 , I have replaced old link to new link

old link 
https://go.microsoft.com/fwlink/p/?LinkId=618719
new link
https://www.kingston.com/us/support/technical/products?model=dtws#DTWS


illfated:
To me, this does not look like the link reported as the invalid link in issue ticket #5562

Both of the links involved in the current PR content change resolve to the same page:
- https://go.microsoft.com/fwlink/p/?LinkId=618719 (resolves to the one below)
- https://www.kingston.com/us/support/technical/products?model=dtws#DTWS

Dansimp:
I would prefer to keep the fwlink as-is (this allows us to change the URL target easily.) The broken link reported in the issue is https://www.kingston.com/us/usb/bootable/ikw700. Which I don't see on this page. Perhaps it's a loc issue. Will need to investigate. But for now, closing this PR.

PR #5616, opened 2019-12-05
Add real scenario for AAD Registered devices

juntakata:
Some of our customers are asking if a convenience PIN is supported on Azure AD Registered devices. As the original document does not mention to AADR devices explicitly, we received similar support requests from these customers. I added an example behavior observed.


mapalko:
An azure ad registered account should not be blocked from provisioning a convenience PIN if the AAD account is added to a local account or a MSA account. The convenience PIN is tied to the unlock credential on the device. You should be able to provision a key for the Azure AS registered account with convenience PIN being set up on the local account.

PIN change should also not be blocked in these scenarios.

e0i:
@juntakata In our understanding the document author's comment contradicts the thesis of the pull request and thus it is being closed. Feel free to re-open the PR for review if that is not the case. Thank you.

PR #5638, opened 2019-12-10
`copy-paste`ability

CaringDev:
update to remove – (en-dash) in code


illfated:
Neat update. Would you be open to a suggestion for a MarkDown Note format improvement just below the text you have modified already?

PR #5676, opened 2019-12-12
Connect-MSGraph should be used

modir:
The other command does not exist anymore in the current version of the module. Connect-MSGraph needs to be used instead. If you run the command mentioned afterwards "Get-AutopilotProfile" then you even get in the error message the sentence "please connect first with Connect-MSGraph".


illfated:
CC: @greg-lindsay

PR #5685, opened 2019-12-13
Release notes link changes to HoloLens section

evmill:
Link for HoloLens release notes was pointing to Mixed Reality docs instead of HoloLens. Redirected.
@scooley


scooley:
@Dansimp this is a good change we should merge :)

PR #5713, opened 2019-12-18
Edit table of default values

cchapin2020:
Edited table of default settings based on Windows Server 2016 domain defaults.  The right on a domain controller is set in the local security policy and is not defined in the Default Domain Controller policy.


illfated:
Well, @Dansimp is credited both as the author and doc manager here, so I presume he will follow up on this when he gets time to do so. In the meanwhile, we could have a copy review from @JohanFreelancer9 just to be clear that we have followed protocol before signing off on this PR.

illfated:
#sign-off

PR #5721, opened 2019-12-19
Updated deferral period for quality updates

silvermarkg:
Deferral period for quality updates was incorrect. Document showed 35 days when deferral is for a maximum of 30 days.


illfated:
Oh, I see. Someone may have confused the number of days for Quality Updates deferral with the general "pause updates" deferral time. Good point.

PR #5749, opened 2019-12-26
Update the group policy path

appcompatguy:
The path in local group policy is now "Computer Configuration\Administrative Templates\System\Device Guard" (at least in 10.0.19041), updating docs to match what is in local group policy.


illfated:
#sign-off

Issue #5756, opened 2019-12-28
Some comments about this documentation page


kennethvs:
Hi,

Just executed a firmware update of my Surface docking, here are my comments on this documentation page, which could use some rewriting if you ask me.

1. You need to download an .MSI from the Microsoft Surface Dock Firmware Update page. The MSI installs and asks for a reboot? Later I read that a reboot is not really necessary.
2. The documentation states:"After you connect Surface Dock to your Surface device, the tool checks the firmware status while running in the background." - does that mean that the MSI must be installed while the Surface Dock is disconnected? or do we mean that we need to disconnect and connect the Surface Dock in order for "something" to start? - and why is there no visual window explaining what is going on?

So, apparently the firmware was updated, but again no visual indication of what happened. or will it only be completed once the power disconnects and it boots again? 

About Manual Installation:
Reconnect your Surface Dock for 2 minutes and then disconnect it from your device. The DisplayPort firmware update will be installed while the hardware is disconnected. The LED in the Ethernet port of the dock will blink while the update is in progress. Please wait until the LED stops blinking before you unplug your Surface Dock from power.

a) what is the difference here compared to the first installation...was that the automatic installation or something? and let me see if I understand this completely - I need to reconnect my surface dock for 2 minutes (so, disconnect if already connected and then connect), within those 2 minutes "something" happens. Then I need to disconnect for some installation to take place, and then when the Ethernet LED stops blinking I need to unplug the Surface Dock from power? 

Manually installing the MSI file may prompt you to restart Surface; however, restarting is optional and not required. -&lt; then remove this from the MSI, its confusing and unnecessary

You will need to disconnect and reconnect the dock twice before the update fully completes. &lt;- this was neither mentioned in the first part and manual installation part

Network Deployment: Msiexec.exe /i /quiet /norestart will do nothing, you must specify the .MSI file after the /i parameter. when people execute this command line nothing will happen.

it says "A log file is not created by default. In order to create a log file, you will need to append "/l*v [path]" two times? why?

For Intune deployment, please specify how the detection method must be created, later the registry locations are explained but nowhere that these could be used as a detection method as well, also here the .msi file must be specified after /i 

[...]

PR #5759, opened 2019-12-28
removed the sentence which appeared twice

RAJU2529:
as per user report  #5756 , i removed the following  sentence

> [!NOTE]
> A log file is not created by default. In order to create a log file, you will need to append "/l*v [path]"


illfated:
#sign-off

Issue #5708, opened 2019-12-17
Check onboarding status


gator55:
1 if the device is not yet onboarded

There should be "0"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2f067243-d029-ec35-cb1d-cc271b00d0a1
* Version Independent ID: 121d2480-5bbc-9a8e-68a6-17f708dc21d4
* Content: [JAMF-based deployment for Microsoft Defender ATP for Mac - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/mac-install-with-jamf#check-onboarding-status)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/mac-install-with-jamf.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/mac-install-with-jamf.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

mypil:
@gator55 -  Thank you for submitting feedback.

From our understanding, the issue has been resolved based on this merged commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/f931514be95fc8ba7d6ac57aabded0685168f3e2. 


Thank you for your contribution to make the docs better! Much appreciated!

PR #5778, opened 2020-01-01
Output was Wrong

joinimran:
it should be 0 if the device is not yet onboarded. Made the correction.

Problem: https://github.com/MicrosoftDocs/windows-itpro-docs/issues/5708


mypil:
@Dansimp - PR has been copy edited and is ready for final review. This is based on issue #5708

Thank you.

cc: @kenwith

Issue #5799, opened 2020-01-06
Blog post link 


czrik:
The URL to Y0av's blogpost is dead. Looks like the article has been taken down

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 69882e91-6bef-d524-7059-20ba67129f4e
* Version Independent ID: ee0f2eb3-97da-4306-5cd1-2b6ae1858ed9
* Content: [Configure Surface Hub Start menu - Surface Hub](https://docs.microsoft.com/en-us/surface-hub/surface-hub-start-menu#feedback)
* Content Source: [devices/surface-hub/surface-hub-start-menu.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/devices/surface-hub/surface-hub-start-menu.md)
* Product: **surface-hub**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

RAJU2529:
I create PR to remove invalid link and I add a similar corresponding information link

greg-lindsay:
The link will be deleted for now. A new blog link is pending.

PR #5800, opened 2020-01-06
i removed broken link and added similar link

RAJU2529:
as per user report #5799 , so i removed the following link
**https://blogs.technet.microsoft.com/y0av/2018/02/13/47/**

and added  Customize and export Start layout link
**https://docs.microsoft.com/en-us/windows/configuration/customize-and-export-start-layout**

### DIFF
--- a/devices/surface-hub/surface-hub-start-menu.md
+++ b/devices/surface-hub/surface-hub-start-menu.md
@@ -185,4 +185,4 @@ This example shows a link to a website and a link to a .pdf file. The secondary 
 
 ## More information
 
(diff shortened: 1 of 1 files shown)


greg-lindsay:
This isn't quite the right fix.  The blog link (and the page) is for surface hub, not Windows 10.  The two are quite different.  We cannot use a link to a Windows 10 article for the Surface Hub article. ;)

greg-lindsay:
I had already started a thread with the author of the old blog post. They will have an updated link by the end of this week. In the meantime I am deleting the current link.

Issue #5832, opened 2020-01-11
Doc does not make sense


termnews:
These instructions don't make any sense, previous comments deliver the details on this statement, but no one assigned to this doc seems to be able to understand these comments. So I just post this to increase the counter -- perhaps Microsoft assigns someone who does more than copy-pasting friendly reply snippets to the issue when there are enough posts.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2a7b356e-3dbd-7f59-845e-a92e05865a52
* Version Independent ID: 711c5497-373f-662c-7fae-ba60f046e50c
* Content: [Microsoft Surface Dock Firmware Update - Surface](https://docs.microsoft.com/en-us/surface/surface-dock-firmware-update#feedback)
* Content Source: [devices/surface/surface-dock-firmware-update.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/devices/surface/surface-dock-firmware-update.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

RAJU2529:
@e0i . Please assign the  author of this article, So that author can directly  submit  comments

illfated:
To quote jvsam: Or is there a way that we can engage the Product Engineering team to help clarify the feedback?

e0i:
Hello, @termnews Would you clarify the problematic part(s) about the document? Is it the snippets? We would like to follow up with clarifying updates. Thanks you for providing feedback.

illfated:
I had to re-read the issue post a couple of times to see the relation, but this is mostly a follow-up response to the content in issue ticket #5756 [**Some comments about this documentation page**](https://github.com/MicrosoftDocs/windows-itpro-docs/issues/5756) and partially #5214 (Better describe the update status / better logging) , which might be closed soon.

Based on that, I presume this is about the uncertainties around the Microsoft Surface firmware update procedure, which may or may not work as intended & expected, depending on how you read the documentation.

greg-lindsay:
I should probably be the listed author of this article (although I didn't write it) because I've taken over managing these topics.  I will address the issue with the product team and try to rewrite the article, or get it rewritten and then edit and republish.

e0i:
Apologies to all for the delayed follow-up.

[...]

Issue #5214, opened 2019-10-18
Better describe the update status / better logging


DennisGaida:
I was happy with the legacy Surface Dock Updater. You start an exe, you see progress. Logging is done in Event Log. I totally don't understand the new Updater. Please don't take this as a rant, but from a user that actually wants to update the dock and fails due to lack of documentation and progress bars. So here are my concerns:

Under "**_To Run Surface Dock Firmware Update_**" it says:
_2. After you connect Surface Dock to your Surface device, the tool checks the firmware status while running in the background_

Please document how I can check that the software is doing the checking. "While running in the background" means that a new Windows Service is installed that is running in the background? When I use the msi logging command documented further below - is that how I can check if it's doing its job? The event log seems to be deprecated.

_3. After several seconds, disconnect your Surface Dock from your device and then wait for 5 seconds before reconnecting. The Surface Dock Firmware Update will normally update the dock silently in background after you disconnect from the dock and reconnect. The process can take a few minutes to complete and will continue even if interrupted._

"After several seconds". What does that mean? How do I know it is finished? There is no progress bar or anything. Same goes for "a few minutes to complete" - how do I know it is complete? I guess the flashing LED in the ethernet port? Is there a signal like multiple flashes? No flashes?

Heading "**_Manual installation_**"
I am supposed to reconnect "for 2 minutes". Do I need to time this? Why 2 minutes? Is there some progress bar somewhere I can check that whatever did it's thing?

Heading "**_Event logging_**"
Since the current version doesn't log to the event log, shouldn't the heading be "logging"?
Why are the event log IDs documented here if only relevant for the legacy versions? I suggest removing everything legacy, and putting that under a "Legacy" heading.

I currently have Component10CurrentFwVersion = 4a26d70, Component20CurrentFwVersion=4a1d570. Display port version looks good, MCU seems "different". Now I have no clue what to do and the documentation doesn't help. I'll do some more of the troubleshooting: Uninstalling the update and randomly unplugging/pluggin back in the dock, but that doesn't seem like a smart process. ([Actually it is the same in this unanswered comment](https://github.com/MicrosoftDocs/windows-itpro-docs/issues/5046#issuecomment-535937604))

I would love a step-by-step documentation. First do this, than that. All the IT-Pro stuff (network deployment & Intune) needs to be way lower since that won't be the primary use case. I can gladly help improving the documentation with a pull request but right now I don't actually understand the process that needs to happen.

---
#### Document Details

[...]

PR #5839, opened 2020-01-13
Update microsoft-surface-brightness-control.md

Cloud-Writer:
This update is a high pri request from Tod Edwards, Sr. Technical Advisor for Devices. Here is what he requested in the work item he filed with the CSS Content team:
-----------------------------------------------------------------------------------------------------------------------------------------------------
Update https://docs.microsoft.com/en-us/surface/microsoft-surface-brightness-control

Version 1.16.137 uses this registry location now:
HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Surface\Surface Brightness Control

We need to keep information on the page that says previous versions of the tool use: 
Computer\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Surface\Surface Brightness Control

If customer doesn't know which version they have, they will know which place they need to edit based on if the registry path exists or not, since the registry location is created when the app is installed.

### DIFF
--- a/devices/surface/microsoft-surface-brightness-control.md
+++ b/devices/surface/microsoft-surface-brightness-control.md
@@ -46,9 +46,14 @@ documentation](https://docs.microsoft.com/windows/desktop/sysinfo/registry).
 1.  Run regedit from a command prompt to open the Windows Registry
     Editor.
     
-      - Computer\HKEY\_LOCAL\_MACHINE\SOFTWARE\Microsoft\Surface\Surface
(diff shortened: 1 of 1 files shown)


illfated:
(ref. https://github.com/MicrosoftDocs/windows-itpro-docs/issues/5832)

CoveMiner:
I installed it and it shows up correctly per today's revision. After a modifying a reg setting, users, do need to restart their machine for the settings to take effect. This is standard for registry modifications but I'll add this info to the article.

illfated:
(ref. #5839, #5756, #5214 )

PR #5913, opened 2020-01-23
Change language to match setting

gcmlpfaltaher:
the note previously stated "Waiting for other disk encryption"; however, the setting in Intune is called "Warning for other disk encryption"


e0i:
@Dansimp This PR is ready to be merged as it has been approved by the document author. Thank you.

PR #5926, opened 2020-01-24
Updating Bloom to Start Gesture on Kiosk page

evmill:
@scooley @yannisle 
Updated the terms at top of doc to include new terms for HL2.
Doc also seems suspect of need of overhaul of instruction on need for XML files with Intune but I want to confirm before making those edits.


scooley:
@DaniHalfin - this is ready to merge :)

PR #5938, opened 2020-01-27
Update user-roles.md

amirsc3:
It should be mentioned that we need an Azure AD Security group. We have seen cases where customers used Office group by mistake and RBAC did not work because of that.


illfated:
Good point.

PR #5984, opened 2020-02-04
Update vpnv2-profile-xsd.md

msarcletti:
"MachineMethod>Eap</MachineMethod" should’t be in the example, because

1.  MachineMethod can only be Certificate 
------ see https://docs.microsoft.com/en-gb/windows/client-management/mdm/vpnv2-csp 
........................VPNv2/ProfileName/NativeProfile/Authentication/MachineMethod
                         This is only supported in IKEv2.
                                This value can be one of the following:
                                •	Certificate

2. A profile conatins either UserMethod or MachineMethod but not both


e0i:
@Dansimp The PR that has been approved by the document author can be merged. Thanks.

Issue #3997, opened 2019-06-11
AMD and MBEC


sylveon:
This page states that HVCI works on both Intel and AMD platforms, however it also states that MBEC is required, which is not present on any AMD CPU ("HVCI works with [...] AMD. CPU new feature is required Mode based execution control (MBE) Virtualization.")

Does that means that it uses the software emulation mentioned later (does it have an impact on performance?) or there's more magic CPU features on AMD that allows it to work? In this case, what would be the AMD equivalent to a 7th generation Intel Core CPU?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1526c4cb-4584-7fc6-e633-a48e13dad7e8
* Version Independent ID: 13a88ed8-f43d-f658-cbe4-d148d5ffaa12
* Content: [Enable virtualization-based protection of code integrity](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-exploit-guard/enable-virtualization-based-protection-of-code-integrity#feedback)
* Content Source: [windows/security/threat-protection/windows-defender-exploit-guard/enable-virtualization-based-protection-of-code-integrity.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/security/threat-protection/windows-defender-exploit-guard/enable-virtualization-based-protection-of-code-integrity.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @levinec
* Microsoft Alias: **ellevin**

e0i:
@officedocsbot assign @e0i

RAJU2529:
@sylveon. processor requirements are mentioned in this website.
**https://docs.microsoft.com/en-us/windows-hardware/design/device-experiences/oem-highly-secure**

or 

**https://www.techrepublic.com/article/microsofts-new-hardwarefirmware-security-standards-promise-highly-secure-windows-10-devices/**

or

**https://www.bleepingcomputer.com/news/security/microsoft-releases-standards-for-highly-secure-windows-10-devices/**

sylveon:
The official page says nothing about MBEC on AMD, the Tweet referred on BleepingComputer was talking exclusively about Intel CPUs, and while TechRepublic says I should have it, Windows reports I don't on my Ryzen 2600X system. Those articles only contributed to the confusion further.

![](https://cdn.discordapp.com/attachments/314403600103112705/586350183479246849/unknown.png)

[...]

PR #6003, opened 2020-02-06
Update notice about MBEC support

sylveon:
This updates the document with findings from #3997, and makes the text clearer.


illfated:
Ref. PR **#6008** (**Updated "Note" about HVCI**)

PR #6004, opened 2020-02-06
Update white-glove.md

daniel-microsoft:
Need to be clear that it is either LOB or Win32, not both.


daniel-microsoft:
Can we move this to the Private repository? I just noticed that I did the commit to the public one.

e0i:
@daniel-microsoft:

> Can we move this to the Private repository? I just noticed that I did the commit to the public one.

cc: @greg-lindsay

PR #6010, opened 2020-02-07
WDAC cmdlets are available on all SKUs for 1903+ builds

isbrahm:
The SKU check which restricted WDAC cmdlets was removed in 1903, so policies can now be crafted on non-Enterprise systems


isbrahm:
@illfated @denisebmsft could you merge these changes? Thanks!

illfated:
#sign-off

PR #6022, opened 2020-02-09
Fix Defender Operational event log name

nathanmcnulty:
Log name does not have a - between Windows and Defender. Updating would help those copying the event log name into an event forwarder or searching a SIEM.


illfated:
I concur. The spelling can also be checked on Windows 10 Home, as the Event Viewer path is there too, even without applying any filter:
```
Log Name: Microsoft-Windows-Windows Defender/Operational
```

PR #6071, opened 2020-02-17
Fixed scripting issue

erroltuparker:
Merge-CIPolicy command had a mistype causing the line to fail


illfated:
Good point. That equal sign looks like it came out of nowhere, especially obvious when comparing with the previous PowerShell script lines on this page.

PR #6104, opened 2020-02-21
Merging update to WDAC types of rules to create. 

jgeurten:


illfated:
Rich diff, for easier overview of the effective changes:
- https://github.com/MicrosoftDocs/windows-itpro-docs/pull/6104/files?utf8=%E2%9C%93&diff=unified&w=1&short_path=73ba7bc#diff-73ba7bc775e3fc96a5a2cf839ff12e42

windows/security/threat-protection/windows-defender-application-control/select-types-of-rules-to-create.md
```diff
@@ -120,9 +120,9 @@ There is a defined list of SIDs which WDAC recognizes as admins. If a filepath a
...
+Wildcards can be used at the beginning or end of a path rule: only one wildcard is allowed per path rule. Wildcards placed at the end of a path authorize all files in that path and its subdirectories recursively (ex. C:\\* would include C:\foo\\* ). Wildcards placed at the beginning of a path will allow the exact specified filename under any path (ex. \*\bar.exe would allow C:\bar.exe and C:\foo\bar.exe). Wildcards in the middle of a path are not supported (ex. C:\\*\foo.exe). Without a wildcard, the rule will allow only a specific file (ex. C:\foo\bar.exe). <br> The use of macros are also supported and useful in the scenario where the OS is installed elsewhere than the C:\ drive. Supported macros: %OSDRIVE%, %WINDIR%, %SYSTEM32%.
```
illfated: The verb points back to "The use of", not to "macros", so it should be in singular form: 
`The use of macros is also supported and useful in [...]`
```suggestion
Wildcards can be used at the beginning or end of a path rule; only one wildcard is allowed per path rule. Wildcards placed at the end of a path authorize all files in that path and its subdirectories recursively (ex. C:\\* would include C:\foo\\* ). Wildcards placed at the beginning of a path will allow the exact specified filename under any path (ex. \*\bar.exe would allow C:\bar.exe and C:\foo\bar.exe). Wildcards in the middle of a path are not supported (ex. C:\\*\foo.exe). Without a wildcard, the rule will allow only a specific file (ex. C:\foo\bar.exe). <br> The use of macros is also supported and useful in scenarios where the system drive is different from the C:\ drive. Supported macros: %OSDRIVE%, %WINDIR%, %SYSTEM32%.
```
(I also adjusted the syntax for improved readability and logical understanding.)

illfated: Looks like I might have to reconsider suggesting that semicolon instead of the colon there, because GitHub thinks it is a comment indicator (as it often is in some programming languages).

illfated:
Good choice using backticks, that is a useful way to block some unexpected changes from machine translation.

PR #6144, opened 2020-03-01
Unite the bold text of a selected option

homoSapiensSapiens:
This is a single option in the configuration and splitting the bold text sections makes it look like there are 3 different options.
image


illfated:
#sign-off

PR #6155, opened 2020-03-03
Update troubleshooting.md

### LINKED ISSUE #6150: Issue with malformed hardware hash - bad length
This issue occurred with a Lenovo ThinkCentere M93p, this might matter. Ideally I'd like to just make a PR for this to document the troubleshooting and resolution steps but it seems a weird fit for any of the autopilot docs.

Pulling the hash with the script worked as expected. Importing the hash into Intune didn't though - selecting the file is fine, it passes the row check, but clicking Import does nothing. A network trace reveals that the import fails with error 400, the error body is  **Cannot convert the literal '[DEVICEHASH]' to the expected type 'Edm.Binary'**

Trying to decode the base64 with this snippet throws **Invalid length for a Base-64 char array or string**
```powershell
[System.Text.Encoding]::ascii.getstring( [System.Convert]::FromBase64String("DeviceHash"))
```

running it through **oa3tool /decodehwhash=DeviceHash** throws **The value specified in /DecodeHwHash is neither a file nor a base64-encoded string**

The issue seems to be that the payload needs padding. All those 'A's are effectively empty data, equal to 000000, but this means the data can end up not ending on a clean full byte moving between 6-bits to 8-bits. So we can use = to "pad" the payload with a non-6-bit padding. [Wikipedia](https://en.wikipedia.org/wiki/Base64#Decoding_Base64_with_padding) explains this better.

The padding isn't necessary per Base64, but the implementation can require it. It seems the process that generates the Base64 does so without padding, but autopilot and Powershell's Base64 implementations require padding. 

So now to fix it - you can either add or remove. I fixed by adding. 

If the hash can't be decoded with the powershell above, to the original hash, add an "**=**". If it still fails, add a second. If it still fails, remove both "**=**" and add an "**A**" instead. If it still fails, to the new hash with an extra "**A**", add "**=**", and repeat until it decodes to mostly illegible text (it stops throwing errors). This is a valid padded base64 string. Replace the hash in the csv with this new one then try again.

To do it by removal, replace the last "**A**" with "**==**". Then try removing an "**=**". Then the other. Then (with both "**=**" removed), replace the next last "**A**" with "**==**", so on, so forth. 

Never have more than 2 "**=**" in a row, and never have them anywhere else but at the very end of the payload. So no "**===**" and no "**A=A**"

The script itself isn't at fault, this seems to be a discrepency between the data generated and stored in devicehardwaredata in [mdm-devdetail-ext01](https://docs.microsoft.com/en-us/windows/win32/dmwmibridgeprov/mdm-devdetail-ext01) and what intune expects, and it seems to be that the encoder that generates the devicehardwaredata isn't encoding with padding. It might have been fixed in subsequent releases of windows, I don't have the build version documented.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: adcfcc02-d931-66e3-6f4f-c14b936358e6
* Version Independent ID: 8972647b-ba21-b68f-b698-5fb6e9e9f1fd
* Content: [Troubleshooting Windows Autopilot - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/windows-autopilot/troubleshooting)
* Content Source: [windows/deployment/windows-autopilot/troubleshooting.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/windows-autopilot/troubleshooting.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

ShrCaJesmo:
Documenting issue and solution referenced in #6150

Effectively, the hash the device provides is unpadded base64 but autopilot (or, at least, the intune UI) can't parse unpadded base64 if it doesn't align properly. Most of the time the data aligns in a way where that doesn't matter. In rare cases it seems it doesn't align right though, and that's where we need to manually pad the payload for autopilot to accept it.

This snippet could be modified and added to the get-windowsautopilotinfo script to autopad directly in the script, effectively fixing this. But some people might still be using the old version of the script, so it might be good to still document this at least for a while.

```powershell
#Put the hash string in this variable
$hash = ""
#Loop switch
$while = $true
#While the hash is still malformed, keep appending to the hash
while ($while){
    #Try decoding the hash
    try{
        #Convert from Base64 to byte array, but don't return the result
        [System.Convert]::FromBase64String($hash) >> $null
        #If the conversion succeeds, stop the loop
        $while = $false
        }
    #If the base64 can't be decoded
    catch{
        #If there's two consecutive = chars at the end, remove both and add an A
        if(($hash.ToCharArray()[-1]-eq '=') -and ($hash.ToCharArray()[-2]-eq '=')){
            #Replace = with nothing and overwrite with the result
            $hash=$hash.replace("=","")
            #Append A
            $hash=($hash+"A")
        }
        #Otherwise, add an =
        else{
            #Append =
            $hash=($hash+"=")
        }
    }
}
#Return the fixed hash
$hash
```

Closes #6150

### DIFF
--- a/windows/deployment/windows-autopilot/troubleshooting.md
+++ b/windows/deployment/windows-autopilot/troubleshooting.md
@@ -42,6 +42,46 @@ For troubleshooting, key activities to perform are:
 - Azure AD join issues.  Was the device able to join Azure Active Directory?
 - MDM enrollment issues.  Was the device able to enroll in Microsoft Intune (or an equivalent MDM service)?
 
+## Troubleshooting Autopilot Device Import
+
+### Clicking Import after selecting CSV does nothing, '400' error appears in network trace with error body **"Cannot convert the literal '[DEVICEHASH]' to the expected type 'Edm.Binary'"**
+
+This error points to the device hash being incorrectly formatted. This could be caused by anything that corrupts the collected hash, but one possibility is that the hash itself, even if completely valid, fails to be decoded.
+
+The device hash is Base64. At the device level, it's encoded as unpadded Base64, but Autopilot expects padded Base64. In most cases, it seems the payload lines up to not require padding, so the process works, but sometimes it doesn't line up cleanly and padding is necessary. This is when you get the error above. Powershell's Base64 decoder also expects padded Base64, so we can use that to validate that the hash is properly padded.
+
+The "A" characters at the end of the hash are effectively empty data - Each character in Base64 is 6 bits, A in Base64 is 6 bits equal to 0. Deleting or adding "A"s at the end doesn't change the actual payload data.
+
+To fix this, we'll need to modify the hash, then test the new value, until powershell succeeds in decoding the hash. The result is mostly illegible, this is fine - we're just looking for it to not throw the error "Invalid length for a Base-64 char array or string". 
+
+To test the base64, you can use the following:
+```powershell
+[System.Text.Encoding]::ascii.getstring( [System.Convert]::FromBase64String("DEVICE HASH"))
+```
+
+So, as an example (this is not a device hash, but it's misaligned unpadded Base64 so it's good for testing):
+```powershell
+[System.Text.Encoding]::ascii.getstring( [System.Convert]::FromBase64String("Q29udG9zbwAAA"))
+```
+
+Now for the padding rules. The padding character is "=". The padding character can only be at the end of the hash, and there can only be a maximum of 2 padding characters. Here's the basic logic.
+
+- Does decoding the hash fail?
+ - Yes: Are the last two characters "="?
+  - Yes: Replace both "=" with a single "A" character, then try again
+  - No: Add another "=" character at the end, then try again
+ - No: That hash is valid
+
+Looping the logic above on the previous example hash, we get the following permutations:
+- Q29udG9zbwAAA
+- Q29udG9zbwAAA=
+- Q29udG9zbwAAA==
+- Q29udG9zbwAAAA
+- Q29udG9zbwAAAA=
+- **Q29udG9zbwAAAA==** (This one has valid padding)
+
+Replace the collected hash with this new padded hash then try to import again.
+
 ## Troubleshooting Autopilot OOBE issues
 
 If the expected Autopilot behavior does not occur during the out-of-box experience (OOBE), it is useful to see whether the device received an Autopilot profile and what settings that profile contained.  Depending on the Windows 10 release, there are different mechanisms available to do that.


illfated:
@ShrCaJesmo : Good job, well done. I did not even get time to read through it before it got merged. Anyway, I trust that the research you posted in the issue ticket was as good a foundation as any, so my hat off to you for pulling through on this one. Thank you for the PR and the ticket you based this on. 👍

ShrCaJesmo:
Happy to help! It was a fun one to figure out. Thanks for your assistance with the process!

ShrCaJesmo:
That said, I just realized that the "decision tree" I tried making lost it's structure, so the actual steps to manually remediate the hash are a bit hard to read

It should have looked like this:
- Does decoding the hash fail?
  - Yes: Are the last two characters "="?
     - Yes: Replace both "=" with a single "A" character, then try again
     - No: Add another "=" character at the end, then try again
  - No: That hash is valid

illfated:
OK, very well. Thanks for keeping up with the content and not letting it go. Feel free to create a new PR to resolve that bit (and maybe reference it with "Ref. #6155 " instead of the Close/Fix method in the new PR).

illfated:
BTW, not sure if you have located a way to update your local public branch to create a new PR branch from. Myself, I use a client tool, Git Extensions, to update my local clone on my computer.

ShrCaJesmo:
I'm not sure I'm following - do you mean like pulling the current state of the https://github.com/MicrosoftDocs/windows-itpro-docs into my fork of it so it's up to date? I've figured out how to make a PR in the other direction to merge the upstream changes, but that's the best I've got - i imagine my fork's pretty outdated at this point. My access to github is pretty limited in general, so I'm limited to the browser without extensions.

illfated:
OK, fair enough, it does not seem to be very important in this case. My comment is probably based on how overzealous I tend to be regarding my own PRs, but as long as you don't get any file conflict, you are basically good to go as is. (Ref. #6156 ) Just as an F.Y.I. I am diagnosed with a mild form of autism (highly functioning), so I often tend to be too focused on details, versus keeping an overview of "the big picture".

ShrCaJesmo:
As someone who tends to overlook the details, thanks for being on the ball! I really appreciate all your help with this!

illfated:
And I appreciate your coding skills (other than my own) and knowing when to spot IT Pro topical stuff and also report them correctly.. :)

PR #6156, opened 2020-03-03
Fixing steps to pad hash - formatting correction

ShrCaJesmo:
Ref. #6155

I didn't realize it right away but I had messed up the format of the bullet list so it was all at the same level, instead of the branching structure it was meant to have - I only realized after it had been merged


illfated:
#sign-off

Issue #6154, opened 2020-03-03
The procedure does not work for ATP Linux Onboarding 


jose-lima2011:
After doing all the steps in the document I got this error

python WindowsDefenderATPOnboarding.py
Generating /Library/Application Support/Microsoft/Defender/com.microsoft.wdav.atp.plist ... (You may be required to enter sudos password)
sudo: plutil: command not found

when typing the command mdatp --health

cloudAutomaticSampleSubmission          : true
cloudDiagnosticEnabled                  : true
cloudEnabled                            : true
definitionsUpdated                      : "1583247251971"
definitionsUpdatedMinutesAgo            : 24
definitionsVersion                      : 79924
edrDeviceTags                           : []
edrEarlyPreviewEnabled                  : "disabled"
edrMachineId                            : null
healthy                                 : false
licensed                                : false
logLevel                                : "info"
machineGuid                             : "XXXXXXXXXXXX-8f85-6ff956d7e663"
orgId                                   : null
realTimeProtectionAvailable             : true
realTimeProtectionEnabled               : true
releaseRing                             : "Production"
versionEngine                           : "100.87.66"

I run the connectivity test and everything is fine:

 mdatp --connectivity-test
Running connectivity test
Testing connection with https://cdn.x.cp.wd.microsoft.com/ping ... [OK]
Testing connection with https://eu-cdn.x.cp.wd.microsoft.com/ping ... [OK]
Testing connection with https://wu-cdn.x.cp.wd.microsoft.com/ping ... [OK]

how can I recommend a service that doesn't work for customers? I inform that I tested in all distros of the azure environment and nothing done the onboard does not happen.

the procedure and the troubleshooting is very poor let's improve this !

---
#### Detalhes do documento

⚠ *Não edite esta seção. É necessário para a vinculação do problema do docs.microsoft.com ➟ GitHub.*

* ID: e2197cba-30ac-1749-e25a-6f39c9ee9d20
* Version Independent ID: bcd14f13-ab59-8e69-5c83-42fee6e51f87
* Content: [Implantar manualmente o Microsoft defender ATP para Linux](https://docs.microsoft.com/pt-br/windows/security/threat-protection/microsoft-defender-atp/linux-install-manually#feedback)
[...]

PR #6159, opened 2020-03-04
update linux-install-manually.md

konstruktoid:
Lint, use `apt-key add` on Ubuntu and `yum-utils` on Red Hat.

ref #6154

Signed-off-by: Thomas Sjögren <konstruktoid@users.noreply.github.com>


konstruktoid:
`sudo: plutil: command not found` occurs on both Red Hat and Ubuntu, but `plutil` isn't available and `plistutil` (https://manpages.debian.org/testing/libplist-utils/plistutil.1.en.html) seems to be the alternative.

PR #6161, opened 2020-03-04
Update troubleshooting.md

ShrCaJesmo:
Adds a note not to delete the associated azure object, with remediation steps if it's been deleted.


illfated:
Good job. Please also remember that there are so-called "Note blobs" you can use to highlight the added text, in case you want the reader to be particularly aware of the note, as well as the importance level.
```
> [!NOTE]
>
> [!TIP]
>
> [!IMPORTANT]
```
(there might be one or two I have forgotten, but those above are the essential ones)

ShrCaJesmo:
Thanks for the info! I've used them before but I always forget the syntax, I'll bookmark this as reference. In this case, since this is basically a list of similar details I thought a note block might be excessive, but I'd be happy to submit a new PR that marks it as such if you think it merits one.

illfated:
It is somewhat up to you, depending on what you think the importance of the added text is. The note blobs are colored blobs (as you can see in https://docs.microsoft.com/en-us/windows/deployment/planning/windows-to-go-overview), so you may want to consider how important it is that the reader notices (!) the added text.

Oh right. I forgot the "warning" blob. Not sure if anything else remains.
```
> [!NOTE]
>
> [!TIP]
>
> [!IMPORTANT]
>
> [!WARNING]
>
```

PR #6165, opened 2020-03-04
Update wufb-compliancedeadlines.md

arcarley:
I have updated to clarify that the change is for version 1709 and above *note* for 1903 and above. I have changed all instances of this in the document (or at least tried to). @jaimeo can you approve?


jaimeo:
@arcarley thanks so much for doing this! It looks like the system wants you to sign a CLA (see above) though?

arcarley:
I just did, should be able to go through now. :)

Issue #3829, opened 2019-05-28
Wrong value for registering devices in Intune


hemetz:
Its current 500 devices

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e4775bcf-978a-f992-31e6-9c15333541e1
* Version Independent ID: 662e95c4-8252-9341-e5be-3111cd3a2299
* Content: [Administering Autopilot](https://docs.microsoft.com/en-us/windows/deployment/windows-autopilot/administer#feedback)
* Content Source: [windows/deployment/windows-autopilot/administer.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/master/windows/deployment/windows-autopilot/administer.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greg-lindsay**

martyav:
Just hopped in to doublecheck, and 500 does appear to be the correct value

PR #6192, opened 2020-03-10
re: #3829, Wrong value for registering devices

JimmyJam6972:
#3829 Wrong value for registering devices in Intune

>Its current 500 devices

Our current docs and screenshots agree that the current max is 500; for example, see the link in the table - https://docs.microsoft.com/en-us/intune/enrollment-autopilot - last updated ~6 months after this article, it contains a screenshot listing a max of 500 rows, with one row for each device


greg-lindsay:
https://docs.microsoft.com/en-us/windows/deployment/windows-autopilot/add-devices

The page already says 500 at a time.  This edit was committed in May 2019 and is no longer valid.

PR #6219, opened 2020-03-12
added dg readiness tool link

RAJU2529:
i added  the dg readiness tool website link next to  hardware readiness tool


beerisgood:
Now only the site need to provide the latest version which include the known fixes

PR #6233, opened 2020-03-13
Update upgrade-to-windows-10-with-the-microsoft-deployment-toolkit.md

baardhermansen:
Updated some markdown for Notes.
Changed lists to conform to new standard.
Added one Note markdown, as i interpreted it as being the intention.
Changed to new image extension support.


greg-lindsay:
I'm rejecting this one - I'm sorry.

If there is some new standard to format numbered lists as repeated 1's then I'm going to push back hard on that standard. I want a document that is readable to the author, even if the publishing platform knows how to number the list automatically.

PR #6234, opened 2020-03-13
Update windows-10-deployment-scenarios.md

baardhermansen:
Some tidying up.
Modified Important and Note markdown.
Replaced obsolete URLs.
Updated erroneous URL.


greg-lindsay:
Please don't change URLs based on what you think they should be. And please don't change every instance of "Note:" to an alert.

illfated:
@greg-lindsay : Correct me if I am wrong, but do you prefer to keep the broken links? Or am I misunderstanding the process, that you want these changes split up into 10 separate PRs?

About the Notes: Are you saying that the Note blobs are deprecated in favor of using only single line plain text?

PR #6247, opened 2020-03-15
Update deploying-the-app-v-51-server.md

baardhermansen:
Added markdown for Important and Notes.
Removed extra line shifts.

### DIFF
--- a/mdop/appv-v5/deploying-the-app-v-51-server.md
+++ b/mdop/appv-v5/deploying-the-app-v-51-server.md
@@ -13,36 +13,26 @@ ms.prod: w10
 ms.date: 06/16/2016
 ---
 
-
 # Deploying the App-V 5.1 Server
-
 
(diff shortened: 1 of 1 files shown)


baardhermansen:
> Will be interesting to see whether or not the removal of blank lines at the end will have any effect on the Feedback section on the docs.microsoft.com page.

It shouldn't, should it? Don't recall seeing that many number of line shifts on other, newer, pages. I've only encountered those on older style formatted docs.

illfated:
I know, normally it should not. I just looked at a completely different page, which I also had not checked before the PR update, but found that the page did not have a Feedback section. Maybe I am just overthinking this and perhaps the problem does not exist at all. (Maybe I should have posted a question if that docs.microsoft.com page was supposed to have a Feedback section or not.)

baardhermansen:
A page that can be edited but doesn't have a Feedback section? That's a combination i've never seen before. On pages that are either well and truly old or for some reason editing is not desired, both Edit and Feedback is missing.

illfated:
Thanks for the details. I will keep that in mind and see if I can trace my way back to the PR where I found the page to be missing the Feedback section.

Issue #5899, opened 2020-01-23
Multiple small updates


SwiftOnSecurity:
ISSUE 1.) This sentence doesn't make sense:
"This may result in content being either incorrectly."

ISSUE 2.) Please note this page also applies to Win10 1909.

ISSUE 3.) Please move www.microsoft.com to under go.microsoft.com. Go uses www as an intermediary when forwarding clients, and BOTH need to be allowed.

Also, if more description could be added to some of the URLs that would be appreciated. Thanks for your thankless work on this, I've found it very helpful at my employer.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d22f6a68-3356-c480-1d1c-05dfa475bc25
* Version Independent ID: 1a96efa2-b98d-cea3-cba0-77a698eabcff
* Content: [Connection endpoints for Windows 10 Enterprise, version 1903 - Windows Privacy](https://docs.microsoft.com/en-us/windows/privacy/manage-windows-1903-endpoints)
* Content Source: [windows/privacy/manage-windows-1903-endpoints.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/privacy/manage-windows-1903-endpoints.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @DaniHalfin
* Microsoft Alias: **v-medgar**

DaniHalfin:
@medgarmedgar could you please take a look?
Thanks!

illfated:
> 1. This sentence doesn't make sense: "This may result in content being either incorrectly."

I have looked through the page history here on GitHub. Looks like this issue has been present ever since this page got transferred from the previous location or repository.

https://www.bing.com/search?q=%22This+may+result+in+content+being+either+incorrectly.%22

Search results indicate that other users have deduced the following to be the complete sentence:

> **This may result in content being either incorrectly downloaded or not downloaded at all.**

illfated:
(Ref. #5930 )

e0i:
@SwiftOnSecurity Passing the issue to an independent technical writer for revision. Thank you for providing feedback.

illfated:
(ref. #6331 )

illfated:
Well, that is one door shut for the attempt to say "this page also applies to Win10 1909."

[...]

Issue #5930, opened 2020-01-25
1909 missing


beerisgood:
Where are the infos for build 1909?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 611b03b8-c215-0f34-c739-7f4bc7fcf51e
* Version Independent ID: 695ca891-bf89-82df-a9a8-90e598512c96
* Content: [Windows 10, version 1903, connection endpoints for non-Enterprise editions - Windows Privacy](https://docs.microsoft.com/en-us/windows/privacy/windows-endpoints-1903-non-enterprise-editions#feedback)
* Content Source: [windows/privacy/windows-endpoints-1903-non-enterprise-editions.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/privacy/windows-endpoints-1903-non-enterprise-editions.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @MikeEdgar
* Microsoft Alias: **v-medgar**

RAJU2529:
might be the same as 1903

beerisgood:
Problem still not fixed.

illfated:
Looks like the 1909 page for **Enterprise** is missing too, so there is probably a long list of things to verify before publishing those 2 pages.

See also issue ticket [**Multiple small updates**](https://github.com/MicrosoftDocs/windows-itpro-docs/issues/5899) **#5899** for details.

medgarmedgar:
We didn't publish the docs for 1909 since it was an incremental update to 1903 and no new egress traffic was discovered.

beerisgood:
Fair enough but this should then written somewhere as we users doesn't know the changes.

illfated:
You mean something like "For 1909, see 1903." ?

beerisgood:
> 
> 
> You mean something like "For 1909, see 1903." ?

For example, yes :)

medgarmedgar:
Sorry, I see now that this question is about the lack of 1909 "raw" traffic article and not the Restricted Traffic settings main article. Let me address these various points:

[...]

PR #6331, opened 2020-03-26
added reference to 1909

MaratMussabekov:
#5899

### DIFF
--- a/windows/privacy/manage-windows-1903-endpoints.md
+++ b/windows/privacy/manage-windows-1903-endpoints.md
@@ -19,6 +19,7 @@ ms.date: 5/3/2019
 **Applies to**
 
 - Windows 10 Enterprise, version 1903
+- Windows 10 Enterprise, version 1909
 
 Some Windows components, app, and related services transfer data to Microsoft network endpoints. Some examples include:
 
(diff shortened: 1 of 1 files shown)


MaratMussabekov:
@JohanFreelancer9

illfated:
By the way, is this PR related to issue ticket #5930 (**1909 missing**)? That ticket in itself is more or less related to ticket #5899 .

medgarmedgar:
Note: there was never a 1909 Enterprise (or 1909 non-Enterprise) endpoints article published.  It was a decision made based on the nature of the 1909 release (primarily bug fixes).

medgarmedgar:
Couple of notes:
1. Since we didn't explicitly collect the "raw traffic" endpoints for 1909 I can't say that there is a 100% match between 1903 and 1909.

2. There is also a 1903 "non-enterprise" article with a similar situation of there being no definitive 1909 "raw traffic" results.

Overall, since the 1909 release was mostly bug fixes, but it did include some features as well, the 1903 results will generally map well if not near 100% to 1909.  

Conclusion: 1903 endpoints are a great guideline for what to expect from 1909, but may not be 100% accurate.

Finally, the next release of Windows 10 will have a separate article for both Enterprise and non-Enterprise endpoints.

Dansimp:
rejecting PR per medgarmedgar's comments

Issue #6335, opened 2020-03-26
Windows 10 1909 template link to be added


RAJU2529:
Administrative templates for windows 10 v1909  download link should be added

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ad012591-5c74-f1cd-d3a2-443768e8ceee
* Version Independent ID: 918ea4bc-0849-3745-5d9d-a21a973bcd3b
* Content: [Enroll a Windows 10 device automatically using Group Policy - Windows Client Management](https://docs.microsoft.com/en-gb/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy)
* Content Source: [windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @ManikaDhiman
* Microsoft Alias: **dansimp**

illfated:
#please-close

PR #6336, opened 2020-03-26
Added admx file link of 1909

RAJU2529:
As per user report #6335.
I added the download link for Administrative templates for windows 10 1909.
**https://www.microsoft.com/download/100591**

### DIFF
--- a/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md
+++ b/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md
@@ -245,7 +245,7 @@ To collect Event Viewer logs:
 
 ### Useful Links
 
-- [Windows 10 Administrative Templates for Windows 10 November 2019 Update 1909](https://www.microsoft.com/download/100591)
+- [Windows 10 Administrative Templates for Windows 10 November 2019 Update 1909](https://www.microsoft.com/download/details.aspx?id=100591)
 - [Windows 10 Administrative Templates for Windows 10 May 2019 Update 1903](https://www.microsoft.com/download/details.aspx?id=58495)
 - [Windows 10 Administrative Templates for Windows 10 October 2018 Update 1809](https://www.microsoft.com/download/details.aspx?id=57576)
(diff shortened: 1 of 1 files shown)


RAJU2529:
@illfated. Please review this PR

windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md
```diff
@@ -245,6 +245,7 @@ To collect Event Viewer logs:
 
 ### Useful Links
 
+- [Windows 10 Administrative Templates for Windows 10 November 2019 Update 1909](https://www.microsoft.com/download/100591)
```
illfated: Looks good, nothing to complain about here. Even so, I would like to suggest following the URL style below, for consistency:
```suggestion
- [Windows 10 Administrative Templates for Windows 10 November 2019 Update 1909](https://www.microsoft.com/download/details.aspx?id=100591)
```
(I know, your existing link opens the exact same page. This is more about code style than practical writing.)

illfated:
I suggested the URL format https://www.microsoft.com/download/details.aspx?id=100591
but if my choice is too nitpicky, fair enough. Feel free to skip it, if confirmed by others.

RAJU2529:
@illfated . I have already checked via Android chrome browser. It does not contains details.aspx.

illfated:
OK, very well then. https://www.microsoft.com/en-us/download/details.aspx?id=100591 works just fine for me from Windows 10 and Android (Firefox). Please note that some web server configurations do not allow access to browse individual files in some directories, partly because the details.aspx might be a hidden dynamic file generated by ASP.NET only to run when an ID is provided in the link (hence the `?id=` part of the link.

Never mind, though, it is not very important. We can wait for follow-up by either @denisebmsft or @ManikaDhiman to resolve this question. If the current link format is acceptable, then you may mark my suggestion as resolved to hide it and to clean up this PR comment flow a little bit.

illfated:
@RAJU2529 : Thank you for your PR, for your issue ticket and for following up this topic in general.

@ManikaDhiman : Thank you for resolving & merging this PR.

PR #6364, opened 2020-03-29
fix typo remove mention of ReFS

Karl-WE:
fixed typo in 66

removing ReFS here. It belongs to "removed features" instead
Having it here on the "Features that are no longer under development" might give a reader the impression that ReFS itself is deprecated which is not, ReFS 3.5 just brought new features to it.


illfated:
#sign-off

PR #6405, opened 2020-04-02
Broken link

ramakoni1:
Removing the sentence that has broken link - could not find a suitable replacement  - checked with CSS SME from IE (Louis Shanks) and confirmed that we can remove the sentence with broken link.


denisebmsft:
Thanks @ramakoni1

ramakoni1:
> Thanks @ramakoni1

Great- thanks for taking care of this so quickly!

PR #6447, opened 2020-04-08
reworded retirement dates

cinglis-msft:
Features still in prod, will be removed after COVID situation calms down


cinglis-msft:
Tagging @jaimeo for review

PR #6459, opened 2020-04-10
Updated documentation to reflect that quotes are not supported when u…

JesseEsquivel:
…sing GPO

Quotes are not supported for ASR exclusions, we need to make this clear to our customers, as it is very confusing to them when reading the ADMX template for the setting - because the ADMX template for this setting actually contains double quotes.  Went through this with a customer until we found that quotes are not supported:
https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/23141041


illfated:
Have you considered whether or not it would be useful to separate that added sentence in a colored **Note**, Warning or **important** Note blob below the existing sentence?

JesseEsquivel:
> Have you considered whether or not it would be useful to separate that added sentence in a colored **Note**, Warning or **important** Note blob below the existing sentence?

Its probably best to put it in as a warning.  Let me see about changing it.

illfated:
#sign-off

PR #6463, opened 2020-04-10
Add additional step for AAD app registration

JesseEsquivel:
(jesquive) Added step to point users to create their Azure AD application and assign required permission, as it is not clear when reading this article that you still need to do that.

### DIFF
--- a/windows/security/threat-protection/microsoft-defender-atp/enable-siem-integration.md
+++ b/windows/security/threat-protection/microsoft-defender-atp/enable-siem-integration.md
@@ -67,7 +67,7 @@ Enable security information and event management (SIEM) integration so you can p
    > [!NOTE]
    > You'll need to generate a new Refresh token every 90 days. 
 
-6. Follow the instructions for [creating an Azure AD app registration for MDATP](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/exposed-apis-create-app-webapp) and assign the correct permissions to it to read alerts.
+6. Follow the instructions for [creating an Azure AD app registration for Microsoft Defender ATP](https://docs.microsoft.com/windows/security/threat-protection/microsoft-defender-atp/exposed-apis-create-app-webapp) and assign the correct permissions to it to read alerts.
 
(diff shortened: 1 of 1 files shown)


illfated:
Maybe I am too caught up in details and taking things too literally, but I am used to seeing a more direct reference to the intention of links within the text, unless the link is connected to a link description text.
> Follow the instructions for creating an Azure AD app registration and assigning the correct permissions to it to read alerts.

In my mind, I would have said something like: "Follow the instructions in the linked page to create an Azure AD app registration and assign the correct permissions to read alerts:"

Maybe you find it unnecessary to edit the text, depending on the existing phrasing in this document. I just find it slightly imprecise or lacking in descriptiveness.

Another option would be to incorporate the link with a descriptive text, similar to the **Related topics** section.

illfated:
Thank you. It looks much better and a lot more conventional now.

Please also note that unless the topic manager, product group or developer team require us to use a specific link locale, it is preferred to use language agnostic links as long as they do not break upon removing the locale part (`/en-us/` in this case). I have tested the link both with and without the /en-us/ part, although it does state that there is no translation for my locale, so it shows me the original page language instead. This is also OK and a fallback until pages are made available in other languages via the localization team.

windows/security/threat-protection/microsoft-defender-atp/enable-siem-integration.md
```diff
@@ -67,6 +67,8 @@ Enable security information and event management (SIEM) integration so you can p
    > [!NOTE]
    > You'll need to generate a new Refresh token every 90 days. 
 
+6. Follow the instructions for [creating an Azure AD app registration for MDATP](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/exposed-apis-create-app-webapp) and assign the correct permissions to it to read alerts.
```
illfated, line 70:
```suggestion
6. Follow the instructions for [creating an Azure AD app registration for Microsoft Defender ATP](https://docs.microsoft.com/windows/security/threat-protection/microsoft-defender-atp/exposed-apis-create-app-webapp) and assign the correct permissions to it to read alerts.
```

mjcaparas:
Thanks @illfated - just waiting on one technical reviewer. Then I can merge this.

illfated:
#sign-off

Issue #6423, opened 2020-04-06
Displaying Audit logs for controlled folder access in Centralized location


VAnjan12:
Hello Team,

We have enabled Controlled Folder Access for all our organization computers in Audit mode and wanted to audit which applications are being used by users and blocked by Controlled folder access so that we can add them to exclusions when we enable it completely.

But audit mode is logging events to only local system as per this document. Is there any way to see all these events for all our computers in a centralized location like in Intune portal or MDATP portal?

Thanks in Advance.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2df7c09c-2fbb-5d2e-8084-dd6b23b19674
* Version Independent ID: e004e29c-7bb1-c577-8892-eb616f634ac0
* Content: [See how controlled folder access can help protect files from being changed by malicious apps - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/evaluate-controlled-folder-access#feedback)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/evaluate-controlled-folder-access.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/evaluate-controlled-folder-access.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @levinec
* Microsoft Alias: **ellevin**

RAJU2529:
@VAnjan12. I don't have much knowledge, but below ides might help      If it possible , take the control of all users computers, collect the event logs, then analysis detaily, then exclude apps  for each users , apply through group policy Editor,.

MaratMussabekov:
Hello @VAnjan12,
unfortunately there is no centralized portal like in Intune.
You have the following options:
1. collect the logs from multiple computers using windows event forwarding
2. use [Azure Monitor](https://docs.microsoft.com/azure/azure-monitor/platform/data-sources-windows-events)

Thank you

PR #6481, opened 2020-04-13
added a tip about Windows Event Forwarding

MaratMussabekov:
#6423


MaratMussabekov:
hello @RAJU2529 ,
**which computer should subscribe event forwarding. Sever computer or Clients computers.** - you should configure client computers to send logs to event collector (which is usually a server)
**To view forwarded event logs by server is required any login username nad password** - no additional credentials are required
Thank you

PR #6497, opened 2020-04-14
Edited short descriptions

jdmartinez36:
Edited metadata descriptions below 100 characters.

### DIFF
--- a/windows/deployment/add-store-apps-to-image.md
+++ b/windows/deployment/add-store-apps-to-image.md
@@ -1,6 +1,6 @@
 ---
 title: Add Microsoft Store for Business applications to a Windows 10 image
-description: This topic describes how to add Microsoft Store for Business applications to a Windows 10 image.
+description: This article describes the correct way to add Microsoft Store for Business applications to a Windows 10 image.
 keywords: upgrade, update, windows, windows 10, deploy, store, image, wim
 ms.prod: w10
 ms.mktglfcycl: deploy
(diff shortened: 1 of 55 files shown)


jdmartinez36:
A couple of these articles are showing large differences in the change windows, but only the descriptions and ms.custom tags were changed. Not sure why it's showing up like that.

@DaniHalfin

illfated:
That is quite normal, because some of those files contain CR/LF (Windows) line breaks whereas GitHub files usually contain NewLine (UNIX) line breaks. When those are corrected, either via PRs created here on GitHub or via proper Git offline desktop tools and uploaded to GitHub with correct line breaks for GitHub, the files often contain changes both at the start (metadata section) and at the last line (EOL/EOF) which makes GitHub consider the entire file changed, so it splits the full file content into their respective before (red)/after (green) sections.

For those cases, it is useful to click the Rich Diff view button on the right hand side, to view mostly only the effective changes in the file.

e0i:
I have hit that CRLF problem several times while pushing commits from GNU/Linux. Didn't know that files with mixed line endings within the repository articles was tolerated practice.

illfated:
I can only assume that the decision to launch a grindingly big chore task to clean all files by replacing any existing CR/LF with NewLine has not been taken yet, nor am I sure if it will be. The interactive consequences with all open PRs, both internal and public, would be quite noticeable.

illfated:
Just to reply to your point that you didn't know that files with mixed line endings within the repository articles was tolerated practice.
GitHub is meant to be flexible on many levels, so it is always up to the maintainers & owners of a repository to manage the content.

PR #6511, opened 2020-04-15
WVD Support

anvascon:
MDATP currently does not support WVD multi-user session. This is in the process of being certified.
Making sure this is visible to customers.


illfated:
(Ref. #6417)

Issue #6523, opened 2020-04-17
Configure and update System Center Endpoint Protection clients


sringler:
The important note in the section section #configure-and-update-system-center-endpoint-protection-clients points out: 

"This step is required only if your organization uses System Center Endpoint Protection (SCEP) and you're onboarding Windows Server 2012 R2."

But the required steps mentioned in these section also applies for Windows Server 2008 R2.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 6f4ed2b3-0390-4d53-2450-14995ee8b0b0
* Version Independent ID: 91b5bb3e-55d4-6a8e-c16d-094f15c28ee7
* Content: [Onboard servers to the Microsoft Defender ATP service - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/configure-server-endpoints)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/configure-server-endpoints.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/configure-server-endpoints.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @mjcaparas
* Microsoft Alias: **macapara**

MaratMussabekov:
Hello @sringler ,
totally agree,
also, I think that this "Important" section could be removed at all as this requirement for Windows Server 2008 R2 and Windows Server 2012 R2 is already mentioned in the article, above, where the actions are listed.

Thank you

mypil:
@sringler - Thank you for submitting feedback.

From our understanding, the issue has been resolved based on this merged commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/c564fde252d5abb91e35c486d4793302fc9f3e7b. 


Thank you for your contribution to make the docs better! Much appreciated!

PR #6571, opened 2020-04-25
removed duplicate note

MaratMussabekov:
The same note is available above in the same article, where the actions are listed
#6523


mypil:
@mjcaparas - PR is ready for final review. This is based on issue #6523.

Thank you.

cc: @kenwith

PR #6572, opened 2020-04-25
Update autopilot-reset.md

ErikMoreau:
Changed the Windows key logo to text, the logo doesn't show up in search previews causing CTRL++R to show up there
ex. search for 'Trigger local Windows Autopilot Reset' on Bing


illfated:
(By the way, the same issue is true for translated pages, like German: **STRG + + R**

PR #6589, opened 2020-04-28
exact retirement dates

cinglis-msft:
Now that change freeze is lifted, providing exact retirement dates


cinglis-msft:
Tagging @jaimeo

PR #6621, opened 2020-05-01
Mistake in the Vulnerability section

piaudonn:
It is "Accounts that have the **Log on as a batch job** user right" instead of Accounts that have the **Deny log on as a batch job** user right. Else that does not make a lot of sense.


illfated:
#sign-off

Issue #6700, opened 2020-05-12
Duplicated path


gswish810:
In "The AD DS transaction log files" section, "%windir%\Ntds\EDB*.log" is listed twice.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7c9b6379-e575-9819-eca1-a0952f66b655
* Version Independent ID: 34d81602-ffa2-5be6-67ef-b42b3b6f17ee
* Content: [Configure Windows Defender Antivirus exclusions on Windows Server 2016 or 2019 - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-antivirus/configure-server-exclusions-windows-defender-antivirus#feedback)
* Content Source: [windows/security/threat-protection/windows-defender-antivirus/configure-server-exclusions-windows-defender-antivirus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/windows-defender-antivirus/configure-server-exclusions-windows-defender-antivirus.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

denisebmsft:
Thank you @RAJU2529 and @gswish810. RAJU2529 fixed the things GSWISH810 caught. Changes will render live by this afternoon sometime.

PR #6707, opened 2020-05-12
removed duplicate entry of  %windir%\Ntds\EDB*.log

RAJU2529:
as per user report #6700.  so i removed  duplicate entry of  **%windir%\Ntds\EDB*.log**


e0i:
@denisebmsft Would you be able to take a glance at this PR that introrduces a minor fix? Thank you.

Issue #6808, opened 2020-05-26
Note section is  repeated two times.


RAJU2529:
At the top of the page,Note section is. Mentioned two times , and content is same, so remove one Note section.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: bd200806-fa21-1384-e489-be8fe923c666
* Version Independent ID: 9b68cddb-aa39-5269-9a70-dfe6c2ce1ff7
* Content: [Policy CSP - Browser - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-browser)
* Content Source: [windows/client-management/mdm/policy-csp-browser.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/policy-csp-browser.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

denisebmsft:
Thank you @RAJU2529. The pull request that fixes this has been merged, and you will see the changes on the live site later today. I am going to go ahead and close this out.

PR #6809, opened 2020-05-26
removed  one note section from the two note secition

RAJU2529:
as per the user report #6808 , so i removed one note section from two.
So please verify and approve this PR.
Thanks


windows/client-management/mdm/policy-csp-browser.md
```diff
@@ -19,9 +19,6 @@ ms.localizationpriority: medium
 
 <hr/>
```
illfated, line 21: I think we should remove these 2 lines at the same time. They are not needed now.
```suggestion
```

illfated:
#sign-off

PR #6827, opened 2020-05-27
Update hololens-multiple-users.md

v-jodben:
Line 40 edit - "  - HoloLens 2 users may select the user tile on the top of the Pins panel to sign out the current user." Removed 'Pins panel' with "Start menu"


e0i:
@martyav Seems like this PR can be merged as it has been apporoved by HoloLens content owners. Thank you.

Issue #5543, opened 2019-11-27
Is device credential as a method to enroll into Intune supported?


desmondkung:
Getting conflicting messages here. GPO has an option to allow device credential to be used for MDM enrollment (for clients 1903 and after), and there's a second note to say that "Device credential group policy setting is not supported for enrolling into Microsoft Intune." Which is which?

I'm on 1909 and still getting "MDM Enroll: Server Returned Fault/Code/Subcode/Value=(MessageFormat) Fault/Reason/Text=(Device based token is not supported for enrollment type OnPremiseGroupPolicyCoManaged)."

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ad012591-5c74-f1cd-d3a2-443768e8ceee
* Version Independent ID: 918ea4bc-0849-3745-5d9d-a21a973bcd3b
* Content: [Enroll a Windows 10 device automatically using Group Policy](https://docs.microsoft.com/en-us/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy#feedback)
* Content Source: [windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @ManikaDhiman
* Microsoft Alias: **dansimp**

MrSpritey:
You really need to choose Option: User Credentials, when arranging the GPO. Otherwise you will receive Error 76 in the Eventlog.

AwakeGitten:
Have noticed the same error. Was the registration of shared hybrid devices "By Device" no longer supported after 1903?

mardahl:
They removed the note that said "Device credential group policy setting is not supported for enrolling into Microsoft Intune".
Why do I ask? Why?!
The issue has not been fixed. please avoid using Device Credentials. you are asking for a world of pain.
It will seem to work if you have the Intune Connector installed, but in reality, it's just falling back to User Credentials.

PR #6833, opened 2020-05-28
User credential preferred

kannanb-github:
Even though Device Credential is an option on the GPO, the device credential gives error while auto-enrollment tasks running through the Task Scheduler. To avoid this error we need to choose the User Credential option from the dropdown to auto-enroll the device. The below line has been updated on the document.
5. Click **Enable**, and select **User Credential** from the dropdown **Select Credential Type to Use**, then click **OK**.

### DIFF
--- a/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md
+++ b/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md
@@ -113,7 +113,7 @@ Requirements:
 
     ![MDM autoenrollment policy](images/autoenrollment-policy.png)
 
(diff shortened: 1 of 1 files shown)


kannanb-github:
This PR closes the issue https://github.com/MicrosoftDocs/windows-itpro-docs/issues/5543

illfated:
GitHub pro tip (for future PRs):

if you include the text (example) "Closes #5543" at the end of the commit text (1-2 lines below the commit message/description), the referenced issue ticket will be closed automatically.

You can also use words like "Resolves" or "Fixes" if you prefer any of those instead of "Closes".

The important part is that the "Closes #...." message line is included in the initial PR commit.

PR #6857, opened 2020-06-01
Adding BYO VPN documentation

mtniehaus:
Added additional instructions for the new BYO VPN feature.  This should go live later in June when the feature is available broadly.


mtniehaus:
Ready for review and publishing, feature is going live for all Intune tenants by Monday evening.

PR #6889, opened 2020-06-04
Update .openpublishing.redirection.json

traya1:


traya1:
Corrected a redirect entry

PR #6893, opened 2020-06-05
Updated licensing changes for Windows Servers

computeronix:
Removed outdated licensing information for Windows servers and added proper licensing server information for MDATP based upon PM information and Product Terms as of April 2020.  Confirm with proper channels and review Product Terms to see matching information.


computeronix:
Added an additional licensing comment from Product Terms to make it more visible discussing the maximum number of devices MDATP can be used on per user as this has been a common ask by customers and is buried in the product terms.

PR #6942, opened 2020-06-12
Clarified key trust & 2016 DC's within AD site

rogersoMS:
@mapalko adjusting the text here based off direct customer feedback from National Australia Bank who deployed this week but 'overlooked' this requirement. Clarified Active Directory site, etc


mapalko:
I don't think we should specify at least 1 in parenthesis after adequate. For most real environments that will not be adequate and authentication requests will get dropped. Maybe we could say 1 or more depending on your authentication load.

Dansimp:
Updated per @mapalko's suggestions. LGTM otherwise. Merging...

PR #6947, opened 2020-06-12
minPwdLength > 14 is not supported at this time

ryanries:
minPwdLength > 14 is not supported at this time

### DIFF
--- a/windows/security/threat-protection/security-policy-settings/minimum-password-length.md
+++ b/windows/security/threat-protection/security-policy-settings/minimum-password-length.md
@@ -26,16 +26,16 @@ Describes the best practices, location, values, policy management, and security 
 
 ## Reference
 
-The **Minimum password length** policy setting determines the least number of characters that can make up a password for a user account. You can set a value of between 1 and 20 characters, or you can establish that no password is required by setting the number of characters to 0.
+The **Minimum password length** policy setting determines the least number of characters that can make up a password for a user account. You can set a value of between 1 and 14 characters, or you can establish that no password is required by setting the number of characters to 0.
 
 ### Possible values
 
-- User-specified number of characters between 0 and 20
+- User-specified number of characters between 0 and 14
 - Not defined
 
 ### Best practices
 
-Set Minimum password length to at least a value of 8. If the number of characters is set to 0, no password is required. In most environments, an eight-character password is recommended because it is long enough to provide adequate security and still short enough for users to easily remember. This value will help provide adequate defense against a brute force attack. Adding complexity requirements will help reduce the possibility of a dictionary attack. For more info, see [Password must meet complexity requirements](password-must-meet-complexity-requirements.md).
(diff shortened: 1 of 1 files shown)


illfated:
OK, so looking at this suggested change, may I conclude that the change I made based on the recommendation from Robert Durff was poorly founded?

@MSRobertD : Would you mind commenting on this case, knowing that you requested the current content?

MSRobertD:
@illfated , happy to comment.  Sorry if this wasn't clear in my original feedback / post. 

Although I'm an external consultant, I work embedded with an MS team that puts Windows through security certifications.  Our own testing showed that this GPO does support values >14.  (And that's a good thing, as we wouldn't pass certification without it, which would have generated a different sort of bug / request to the feature team.)  So my original feedback was simply that: hey, this GPO does seem to support > 14, so can we change the documentation to show that?  That's it.  If this is, in fact, a bug in the GPO itself, then let's have a different conversation.  

For an FTE contact to loop in from our team (if needed), please contact Mike Grimm (mike.grimm@microsoft.com).

ryanries:
Hello,

I am also a Microsoft FTE, my alias is jories, and I'm a debug engineer on the Windows debug team. While the Group Policy GUI was updated a couple years ago to allow users to modify the Minimum Password Length in Group Policy and Local Policy to 20, the corresponding changes were never actually made to the SAM (Security Accounts Manager) to actually allow that policy to be successfully applied. In practice, if one attempts to apply a Group Policy where the Minimum Password Length has been configured to > 14 characters, an event from SceCli will be generated in the event log of the machine stating that there was an error applying security policy.

Full support for MinPwdLen > 14 may be revisited again later in the future, but it's not a hot issue for the Feature Team right now, despite many customers asking. It has proven to not be an easy or risk-free change. Changing this restriction is similar to changing MAX_PATH which has been painstaking to change and causes lots of things to break.

illfated:
#sign-off

e0i:
@martyav @Dansimp The PR seems ready to be merged. Thank you.

windows/security/threat-protection/security-policy-settings/minimum-password-length.md
```diff
@@ -26,16 +26,16 @@ Describes the best practices, location, values, policy management, and security
...
+Set Minimum password length to at least a value of 8. If the number of characters is set to 0, no password is required. In most environments, an eight-character password is recommended because it is long enough to provide adequate security and still short enough for users to easily remember. Minimum Password Length of greater than 14 is not supported at this time. This value will help provide adequate defense against a brute force attack. Adding complexity requirements will help reduce the possibility of a dictionary attack. For more info, see [Password must meet complexity requirements](password-must-meet-complexity-requirements.md).
```
illfated, line 38: I suppose you are suggesting this change:
```suggestion
Set Minimum password length to at least a value of 8. If the number of characters is set to 0, no password is required. In most environments, an eight-character password is recommended because it is long enough to provide adequate security and still short enough for users to easily remember. A minimum password length greater than 14 is not supported at this time. This value will help provide adequate defense against a brute force attack. Adding complexity requirements will help reduce the possibility of a dictionary attack. For more info, see [Password must meet complexity requirements](password-must-meet-complexity-requirements.md).
```
The sentence changed is this one (just in case the change does not get highlighted): "Minimum Password Length of greater than 14 is not supported at this time."

PR #6972, opened 2020-06-16
close Final/ tag

ricksentam:


illfated:
I admit that I don't have enough experience with GPO and ADMX files, but is that last SyncML the only place where that uncommon tag needs to be used? There are 3 other SyncML code blocks on the page containing the same `<Final/>` tag as the last one where you request this change. I would think that if one SyncML needs this change, the 3 others should get the same change too, but as I said, I am not an expert. Being somewhat ignorant, I also thought that the existing tag, `<Final/>` is a closing tag, whereas an opening tag would look like this: `<Final>` -- I may very well be mistaken.

Looking forward to some feedback on the details.

ManikaDhiman:
</Final/> is an invalid tag in XML. Final should only look like <Final/>. Closing this PR as no doc update is required.

PR #6996, opened 2020-06-19
Update windows-10-start-layout-options-and-policies.md

warren-msft:
The Server OSs were missing from the applies to list. They are supported and not having them listed will cause some customers to open cases to confirm support.


illfated:
(ref. closed PR #6995 )

illfated:
#sign-off

PR #7000, opened 2020-06-19
Correcting spacing within info boxes

martyav:
I noticed the formatting on this page looked a little odd when I was reading over it for another task.

I think the intent was to have one big info box up top, so I added extra angle brackets on the blank lines between paragraphs

I also inserted a line break and angle bracket in another info box, where it seemed like it was intended


illfated:
🎉 Also, compliments on catching a round number for your PR. 😉

PR #7005, opened 2020-06-21
Adding note about using nltest to check for 2016 DC

rogersoMS:
@mapalko & @Dansimp 

Based on feedback from my customer and assistance from the GTP/AADP identity team - adding the nltest command to find a 2016 DC. This missing information resolved my customer's issue with problematic AADJ devices.


mapalko:
I don't think this makes sense to put in this specific part of the documentation. I think it should be included as a note in the actual deployment guidance.

e0i:
Closing the PR based on Matthew's comment above. Thank you for taking the initiative to provide contributions.

Issue #7030, opened 2020-06-25
Need direct download links


RAJU2529:
Needed download links for edge browser and for adk 2004 , addson pe 2004
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d651de3c-b8eb-e8d9-8138-1eadb758bf2e
* Version Independent ID: a19a8b65-0f4a-efd5-7f83-56bd07cc8c3b
* Content: [What's new in Windows 10, version 2004 - What's new in Windows](https://docs.microsoft.com/en-us/windows/whats-new/whats-new-windows-10-version-2004?ranMID=24542&ranEAID=je6NUbpObpQ&ranSiteID=je6NUbpObpQ-QJM7gBYv2T5Eowt0B2Kprw&epi=je6NUbpObpQ-QJM7gBYv2T5Eowt0B2Kprw&irgwc=1&OCID=AID2000142_aff_7593_1243925&tduid=%28ir__rmtvdnorickft3adxmdf30kppu2xnfsdqwlrvgyc00%29%287593%29%281243925%29%28je6NUbpObpQ-QJM7gBYv2T5Eowt0B2Kprw%29%28%29&irclickid=_rmtvdnorickft3adxmdf30kppu2xnfsdqwlrvgyc00#servicing)
* Content Source: [windows/whats-new/whats-new-windows-10-version-2004.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/whats-new/whats-new-windows-10-version-2004.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

greg-lindsay:
I would prefer not to have direct download links on this page as these links can change, and it's a blind link. It's better to link to the summary page that is updated if the links change.

PR #7031, opened 2020-06-25
added direct downloading links

RAJU2529:
as per the user report #7030 , so i added the following direct downloading links

windows 10 adk v2004
**https://download.microsoft.com/download/8/6/c/86c218f3-4349-4aa5-beba-d05e48bbc286/adk/adksetup.exe**

windows 10   pe addson  v2004
**https://download.microsoft.com/download/3/c/2/3c2b23b2-96a0-452c-b9fd-6df72266e335/adkwinpeaddons/adkwinpesetup.exe**

setupdiag
**https://download.microsoft.com/download/d/8/1/d8149356-6590-4bec-b1bd-a2adcf84ace9/SetupDiag.exe**


greg-lindsay:
I would prefer not to have direct download links on this page as these links can change, and it's a blind link. It's better to link to the summary page that is updated if the links change.

PR #7045, opened 2020-06-27
Revert "replaced word diable to disable"

illfated:
Reverts MicrosoftDocs/windows-itpro-docs#7011

The reason for this revert is pretty simple: The only change provided by the PR #7011 is the address of the internal page link
```html
#connectivity-diableprintingoverhttp">Connectivity/DiablePrintingOverHTTP
```
Because the target text was not changed at the same time, and the MS Docs team consensus is that the issue must be left as it is for now, until someone changes the typo in the operating system itself, this change needs to be left unchanged too.


illfated:
Never mind, this "revert" function did not make any change at all to the target file. Possibly resolved already. /Closing

PR #7050, opened 2020-06-30
Fix typo

icelynjennings:
is has -> has


illfated:
Nice. Good catch. Thank you.

Issue #7021, opened 2020-06-23
Doubt


durbanbr:
Hi,
I would like to understand why some exclusions have two bars, as an example:

% windir% \ SoftwareDistribution \ Datastore **\\** tmp.edb

Also other doubts, as some do not have * before the extension:

% windir% \ Security \ database \ **.chk**

Now two more questions: Are these exceptions the same file names?

% systemdrive% \ System Volume Information \ DFSR \ **$ db_normal $**
% systemdrive% \ System Volume Information \ DFSR \ **$ db_dirty $**
% systemdrive% \ System Volume Information \ DFSR \ **$ db_clean $**
% systemdrive% \ System Volume Information \ DFSR \ **$ db_lostl $**

and this one, is the point right?

% systemroot% \ Sysvol \ Domain \ **.**Scripts.ini

durbanbr:
Excuse me,
I translated the page into Portuguese and it was not so faithful ... But even so I was still in doubt of some things:

Because it has this asterisk here, it means that regardless of the folder it is in, you have to delete this file?
%windir%* \SoftwareDistribution\Datastore\*\Datastore.edb


and keep the doubt of these file names, if it is the name with this symbol $:

% systemdrive% \ System Volume Information \ DFSR \ $ db_normal $
% systemdrive% \ System Volume Information \ DFSR \ $ db_dirty $
% systemdrive% \ System Volume Information \ DFSR \ $ db_clean $
% systemdrive% \ System Volume Information \ DFSR \ $ db_lostl $

RAJU2529:
@durbanbr . Which article are you referring?

durbanbr:
Sorry,
https://docs.microsoft.com/pt-br/windows/security/threat-protection/microsoft-defender-antivirus/configure-server-exclusions-microsoft-defender-antivirus#list-of-automatic-exclusions

durbanbr:
Sorry again, I realized now that the first example didn't paste correctly.

%windir%***** \SoftwareDistribution\Datastore**\*\**tmp.edb

illfated:
@durbanbr : Thank you, both for pointing out the translation issue and asking if the information in the /pt-br/ page is correct or not.

1. The asterisk ( ` * ` ) is frequently used as an indicator that any file name string or data string can be entered there, meaning that some entries use the asterisk ( ` * ` ) to allow for multiple matches to the complete string content.
[...]

PR #7052, opened 2020-06-30
MD-ATP/Exclusions: avoid translation of keywords

illfated:
**Description:**

As recommended by Tina McNaboe (@TinaMcN) in issue ticket #7021 (**Doubt**), I suggest encapsulating all OS strings and keywords in MarkDown back ticks to avoid automatic translation of text that should never be translated, as well as keeping the strings as coherent and unmodified as possible.

To paraphrase:
> While the engineers are working on a fix to avoid over-localization like this on machine translated articles, it may be a few months before it's fixed. In the meantime, it would be helpful to encapsulate the non-localizable strings in the English article.

It is also a common problem that Machine Translation adds spacing where it should not be, even if the surrounding text is kept untranslated. This type of issue tends to be less easy to discover for non-English readers who rely more on the translated pages than on the /en-us/ pages.

This PR aims to reduce these issues, at least for this page, to keep it as correct as possible, even when filtered through Machine Translation.

**Changes proposed:**
- Encapsulate listed strings, filenames and extensions in MD back ticks
- Make sure that the encapsulated strings don't keep extra backslashes

**Additional corrections, since we are already editing this page:**
- Replace TechNet URLs with their permanent docs.microsoft.com redirects
- Replace MSDN URLs with their permanent docs.microsoft.com redirects
- Reduce a docs.microsoft.com link to internal tag link (same page)
- Adjust "Currentcontrolset" to CurrentControlSet (3 occurrences)

**Ticket closure or reference:**

Ref.  #7021 (keep open until changes migrate to localized pages)

### DIFF
--- a/windows/security/threat-protection/microsoft-defender-antivirus/configure-server-exclusions-microsoft-defender-antivirus.md
+++ b/windows/security/threat-protection/microsoft-defender-antivirus/configure-server-exclusions-microsoft-defender-antivirus.md
@@ -356,7 +356,7 @@ This section lists the file type exclusions, folder exclusions, and the process 
 
 This folder is specified in the registry key `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Print\Printers\DefaultSpoolDirectory`
 
-- `%system32%\spool\printers\*
+- `%system32%\spool\printers\*`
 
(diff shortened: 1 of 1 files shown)


illfated:
#sign-off

Issue #6904, opened 2020-06-06
Missing endpoints for WU troubleshooting


gustavo-maia:
[Enter feedback here]
Seems that a few URLs are missing for the required Windows Update endpoints, based on the other documentation at https://docs.microsoft.com/en-us/windows/privacy/manage-windows-1809-endpoints#windows-update

Like the "*delivery.mp.microsoft.com", instead of the mentioned "*.dl.delivery.mp.microsoft.com".

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 349d337e-ca0f-a40a-a522-bd1ae71c7d1e
* Version Independent ID: 50f0c3b0-eaad-89c0-eb3c-878e7564d6f6
* Content: [Windows Update troubleshooting - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/update/windows-update-troubleshooting)
* Content Source: [windows/deployment/update/windows-update-troubleshooting.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/update/windows-update-troubleshooting.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jaimeo
* Microsoft Alias: **greglin**

PR #7077, opened 2020-07-02
add missing URLs

VLG17:
Added missing URLs according to https://docs.microsoft.com/en-us/windows/privacy/manage-windows-1809-endpoints#windows-update

In my opinion some of them can be excluded (for example *.dl.delivery.mp.microsoft.com is already included in *.delivery.mp.microsoft.com), but i guess someone else should make that call. I have added all of them, just to be sure.

https://github.com/MicrosoftDocs/windows-itpro-docs/issues/6904

### DIFF
--- a/windows/deployment/update/windows-update-troubleshooting.md
+++ b/windows/deployment/update/windows-update-troubleshooting.md
@@ -110,7 +110,8 @@ If downloads through a proxy server fail with a 0x80d05001 DO_E_HTTP_BLOCKSIZE_M
 You may choose to apply a rule to permit HTTP RANGE requests for the following URLs: 
 
 *.download.windowsupdate.com  
-*.dl.delivery.mp.microsoft.com  
+*.dl.delivery.mp.microsoft.com
+*.delivery.mp.microsoft.com
 *.emdl.ws.microsoft.com
 
 If you cannot permit RANGE requests, keep in mind that this means you are downloading more content than needed in updates (as delta patching will not work).
@@ -161,6 +162,10 @@ Check that your device can access these Windows Update endpoints:
 - `http://*.download.windowsupdate.com`
 - `http://wustat.windows.com`
 - `http://ntservicepack.microsoft.com`
+- `https://*.prod.do.dsp.mp.microsoft.com`
+- `http://*.dl.delivery.mp.microsoft.com`
+- `https://*.delivery.mp.microsoft.com`
+- `https://tsfe.trafficshaping.dsp.mp.microsoft.com`
  
  Allow these endpoints for future use.
  


jaimeo:
@arcarley can you confirm this edit to the endpoints is correct?

jaimeo:
Still waiting on confirmation from SMEs that this amendment is accurate and complete.

arcarley:
Unfortunately, I do not know. I have emailed the appropriate SMEs for verification and included you, @jaimeo

e0i:
@jaimeo Is there any response from SMEs for this PR? Also, please suggest whether the PR shall be discarded or not if there is no response. Thank you.

jaimeo:
We've heard nothing as far as I know. @arcarley can you advise?

windows/deployment/update/windows-update-troubleshooting.md
```diff
@@ -110,7 +110,8 @@ If downloads through a proxy server fail with a 0x80d05001 DO_E_HTTP_BLOCKSIZE_M
 You may choose to apply a rule to permit HTTP RANGE requests for the following URLs: 
 
 *.download.windowsupdate.com  
-*.dl.delivery.mp.microsoft.com  
+*.dl.delivery.mp.microsoft.com
+*.delivery.mp.microsoft.com
```
narkissit, line 114: *.delivery is redundant (included in *.dl.delivery) so you can exclude it

narkissit: I think it would be best to also add a link to refer to the document that contains the complete list of payload URLs and service URLs. However when it comes to Payload URLS those are the only ones relevant for the "Range" support issue.

windows/deployment/update/windows-update-troubleshooting.md
```diff
@@ -161,6 +162,10 @@ Check that your device can access these Windows Update endpoints:
 - `http://*.download.windowsupdate.com`
 - `http://wustat.windows.com`
 - `http://ntservicepack.microsoft.com`
+- `https://*.prod.do.dsp.mp.microsoft.com`
```
narkissit, line 165: This is a service URL it's not needed in the list to permit RANGE requests (only payload URLs need to be allowed).

narkissit: The list here should match the list above, there are some service URLs / some URLs that are no longer in use here.

windows/deployment/update/windows-update-troubleshooting.md
jaimeo: @narkissit which document would that be (to link to)? Do you mean https://docs.microsoft.com/en-us/windows/deployment/update/waas-delivery-optimization ?

PR #7084, opened 2020-07-02
Update web-content-filtering.md

krupatms:
Changes with removed partner references.


krupatms:
@levinec - please review and let's hold of publishing these until Monday, where the blog will be published as well.

PR #7098, opened 2020-07-07
Fixed Location reg keys

linque1:
Fixed Location reg keys


linque1:
#sign-off

e0i:
cc @denisebmsft

Issue #7129, opened 2020-07-10
MAC and Linux Support


v-rapate:
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 034ffda3-6ef7-13f4-6341-82ac3c11f455
* Version Independent ID: 5527ef3f-6361-b037-d9c3-3d520cf2ef2c
* Content: [Threat & Vulnerability Management supported operating systems and platforms - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/tvm-supported-os)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/tvm-supported-os.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/tvm-supported-os.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @levinec
* Microsoft Alias: **ellevin**

v-rapate:
Microsoft Defender ATP for Linux is now generally available!
https://techcommunity.microsoft.com/t5/microsoft-defender-atp/microsoft-defender-atp-for-linux-is-now-generally-available/ba-p/1482344

Deploy Microsoft Defender ATP for Mac in just a few clicks
https://techcommunity.microsoft.com/t5/microsoft-defender-atp/deploy-microsoft-defender-atp-for-mac-in-just-a-few-clicks/ba-p/1341619

RAJU2529:
@e0i . Assign user to check PR #7133

levinec:
Hi @RAJU2529  and @v-rapate, thanks for your feedback. However, not all Microsoft Defender ATP capabilities have the same supported operating systems/platforms (including threat and vulnerability management). That's why we have one page for Microsoft Defender ATP requirements: https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/minimum-requirements, and a different one for threat and vulnerability management: https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/tvm-supported-os. Currently, Mac and Linux are not supported for threat and vulnerability management. Hope that clears that up!

PR #7133, opened 2020-07-10
added  mac  and linux links

RAJU2529:
as per the user report #7129 .  so i added  ATP  mac  link and ATP linux  link.
Also added operating system support.  
This PR  needs to be verified before merging.


levinec:
Threat and vulnerability management does not currently support Mac and Linux.

Issue #6968, opened 2020-06-15
Please update properly the codes for EventId 14, WS2019 is reporting Credential Guard configuration: 0x2, 0 when CG is running. and is not documented.


carlosmayol:
[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 41c2cb44-269d-3334-019a-726eb5432f1c
* Version Independent ID: 251024cc-b818-18c6-a112-4ac0276cb4b1
* Content: [Manage Windows Defender Credential Guard (Windows 10) - Microsoft 365 Security](https://docs.microsoft.com/en-us/windows/security/identity-protection/credential-guard/credential-guard-manage#enable-windows-defender-credential-guard)
* Content Source: [windows/security/identity-protection/credential-guard/credential-guard-manage.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/identity-protection/credential-guard/credential-guard-manage.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @DulceMontemayor
* Microsoft Alias: **dansimp**

e0i:
Closing via merged PR #7166

PR #7166, opened 2020-07-14
Update credential-guard-manage.md

jirenugo:
Updating documentation to include UEFI event


martyav:
Related to issue #6968, "Please update properly the codes for EventId 14, WS2019 is reporting Credential Guard configuration: 0x2, 0 when CG is running. and is not documented."

denisebmsft (approved):
Thanks!

Issue #6600, opened 2020-04-29
No way to download anything from the Store with WSUS configuration


carloreggiani:
It is still not clear the correlation between WSUS configuration and the ability for clients to download application from our Microsoft Business Store (https://businessstore.microsoft.com): users are still getting error message: "This install is prevented by policy. Ask your admin to enable Windows Update.". Could you add in documentation any clarification about WSUS and Windows Store correlation and configuration?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: fe35c19c-fd29-8405-ad9f-441204812e6b
* Version Independent ID: fc0fe6ce-4b92-3cff-d6fa-c791ba2df466
* Content: [Deploy Windows 10 updates using Windows Server Update Services (Windows 10) - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/update/waas-manage-updates-wsus)
* Content Source: [windows/deployment/update/waas-manage-updates-wsus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/update/waas-manage-updates-wsus.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jaimeo
* Microsoft Alias: **jaimeo**

RAJU2529:
@carloreggiani . Please check the group policy, there is might SRP as setted to disallowed or 
 Policy is modified to run only specific apps..
Check the all registry key under khlm\ software \policies\ machine   hive

VLG17:
@carloreggiani Are you still getting this error? Did it work in the past, or it never worked in your case?

Did you go through the configuration steps described in the article you mentioned? Did you set the intranet Microsoft update service location to your WSUS FQDN?

You can also check [this article](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2008-R2-and-2008/cc720539(v=ws.10)?redirectedfrom=MSDN), which talks about different environment scenarions and different GPO settings which need to be configured in each case. 

My recommendation is to [open a support case](https://support.microsoft.com/en-us/contactus/) in order for a support engineer to look into this issue and the error you are getting and check your configuration. If the conclusion is that we need to add something to the article in terms of configuration, which is currently missing, we will gladly add it.

carloreggiani:
Hi @RAJU2529 and @VLG17: the WSUS policy are fine for Windows Update from our internal WSUS server.

[...]

PR #7899, opened 2020-07-20
Reg key breaks Windows Store connectivity

VLG17:
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/6600

### DIFF
--- a/windows/deployment/update/waas-manage-updates-wsus.md
+++ b/windows/deployment/update/waas-manage-updates-wsus.md
@@ -82,6 +82,9 @@ When using WSUS to manage updates on Windows client devices, start by configurin
 9. Under **Options**, from the **Configure automatic updating** list, select **3 - Auto download and notify for install**, and then click **OK**.
 
    ![Example of UI](images/waas-wsus-fig5.png)
+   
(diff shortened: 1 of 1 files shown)


jaimeo:
@VLG17 i'm a little unclear on your edit. What is "Group Policy Management Editor?" Do you mean "Group Policy Management Console?" And if so, I don't think you can edit regkeys in GMPC. Do you mean to use Registry Editor to check that that key is not enabled?

VLG17:
@jaimeo What i wanted to say was to use gpedit.msc to verify thaat specific key is not enabled.

jaimeo:
@VLG17 maybe I'm just not awake yet, but can you check whether a regkey is enabled in GPMC? Wouldn't you use Regedit.exe for that?

VLG17:
@jaimeo you're right, i think it's me who's not awake. It needs to be checked with Regedit...don't know why i said Group Policy Management Editor, i'll change that to Registry Editor

jaimeo:
@VLG17 no worries! I was thinking "gee did they add some new feature to GPMC?" :^)

Issue #5620, opened 2019-12-06
Powershell cmdlet for cache summary does not work


CharlesBaldridge-MS:
This document states "Starting in Window 10, version 1903, get-DeliveryOptimizationPerfSnap has a new option -CacheSummary which provides a summary of the cache status."  I have attempted to use this switch to verify cache information on both Windows 10 Enterprise x64 1903 and 1909 but the switch does not function on either:  "Get-DeliveryOptimizationPerfSnap : A parameter cannot be found that matches parameter name 'CacheSummary'."

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8d745280-391c-4f6d-6a35-7125fa263548
* Version Independent ID: 0ccd8add-b5fb-469c-509e-e009c43a00ac
* Content: [Set up Delivery Optimization](https://docs.microsoft.com/en-us/windows/deployment/update/waas-delivery-optimization-setup#feedback)
* Content Source: [windows/deployment/update/waas-delivery-optimization-setup.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/update/waas-delivery-optimization-setup.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jaimeo
* Microsoft Alias: **jaimeo**

jaimeo:
@narkissit can you advise?

RAJU2529:
@CharlesBaldridge-MS. Just now i entered the following command on PowerShell with admin rights, even though i received the same error on 1909 64bit.

**Get-DeliveryOptimizationPerfSnap -CacheSummary**


![powershell](https://user-images.githubusercontent.com/3296790/70345525-4b845400-1882-11ea-95bc-68bd61acb120.JPG)

illfated:
Tested the command on both Windows 10 Pro 1903 and Windows 10 Home 1909, both gave me the following output:

```powershell
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

Try the new cross-platform PowerShell https://aka.ms/pscore6

PS C:\WINDOWS\system32> Get-DeliveryOptimizationPerfSnap -CacheSummary
Get-DeliveryOptimizationPerfSnap : A parameter cannot be found that matches parameter name 'CacheSummary'.
At line:1 char:34
+ Get-DeliveryOptimizationPerfSnap -CacheSummary
[...]

PR #7908, opened 2020-07-21
Remove incorrect flag for cmdlet

narkissit:
"CacheSummary" was never added as a flag (may have been a left over in an old spec)


illfated:
(ref. #5620)

Issue #7914, opened 2020-07-21
Explicitly call out the domain controller line of sight requirement for first time use of Windows Hello for Business on Hybrid Azure AD Join


mepples21:
It would be helpful if there was a more explicit callout for the requirement that a remote Hybrid Azure AD Join machine must have line of sight to a domain controller to use the Windows Hello for Business credential that was registered the first time. The diagram in this [section](https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-how-it-works-authentication#hybrid-azure-ad-join-authentication-using-a-key) shows a domain controller in the flow, but it isn't obvious from reading the documentation that this line of sight must be there. This has lead to some confusion with remote deployments of Windows Hello for Business for mostly remote workforces.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 10fcf686-1781-8a62-16bb-b26b3f5982fd
* Version Independent ID: 37d733ad-e809-1674-ffbc-ae3d0c495423
* Content: [How Windows Hello for Business works - Authentication - Microsoft 365 Security](https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-how-it-works-authentication)
* Content Source: [windows/security/identity-protection/hello-for-business/hello-how-it-works-authentication.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/identity-protection/hello-for-business/hello-how-it-works-authentication.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @mapalko
* Microsoft Alias: **mapalko**

joinimran:
Hello @mepples21 - Thank you for your feedback. Based on your feedback we have already added a note as an important section in the document. 
![image](https://user-images.githubusercontent.com/47118050/99191775-aad82d00-2790-11eb-8806-b502d2267c4a.png)

Thanks.
Imran.

illfated:
@e0i : FYI, looks like this one was fairly quickly resolved the same day as the ticket was created.

PR #7916, opened 2020-07-21
Update hello-how-it-works-authentication.md

RavennMSFT:
Added notes to call out remote work related feedback that requires LoS to DC in key-trust and cert-trust as pre reqs for first time logon.


illfated:
(ref. #7914 (Explicitly call out the domain controller line of sight requirement for first time use of Windows Hello for Business on Hybrid Azure AD Join) )

PR #7918, opened 2020-07-22
Wrong screenshot - Request a reseller relationship

rogersoMS:
@brentkendall and @mtniehaus could we please update the screenshot under: "Click the Request a reseller relationship link:"
Text says "Click the Request a reseller relationship link:" however screenshot below has focus on 'Add Customer' tab


mtniehaus:
That's not the way the page works - it's not highlighted, that's exactly what the page looks like:  A button for "Add customer" and a link for "Request a reseller relationship".  So this would need to be a DCR request to the Partner Center team.

greg-lindsay:
This isn't a valid edit for this article. It just breaks the image.

PR #7955, opened 2020-07-24
Update collect-diagnostic-data.md

kurtsarens:
diagnostic data is also part of the investigation package that can be collected via MDATP. Adding note and ref


Dansimp:
Closing this PR as the content is already updated from a different PR.

PR #7966, opened 2020-07-27
Update policy-csp-update.md

chriti:
The TargetRelaseVersion is not correct, this feature was added in 2004 just need the document updated to reflect that.


ManikaDhiman:
Hi @chriti, thanks for the feedback. The TargetRelaseVersion policy setting was added in Windows 10, version 2004 and then backported to version 1803 and later. In the CSP docs, we mention the earliest Windows 10 version in which the policy is available. For the TargetRelaseVersion policy, the doc correctly reflects the version details. I am closing this PR as no doc change is needed. Thanks!

Issue #7106, opened 2020-07-08
Defender AV in Title vs Defender ATP in Applies To Section


aakash-shah:
Hello! The title of the page references "Microsoft Defender Antivirus", but the applies to section explicitly and only references "Microsoft Defender Advanced Threat Protection (Microsoft Defender ATP)". Can you clarify if this document is intended to apply to the native Microsoft Defender Antivirus too? If so it would be helpful to update the Applies to section to clarify that it applies to both "Microsoft Defender Antivirus" and "Microsoft Defender Advanced Threat Protection (Microsoft Defender ATP)" to avoid any confusion.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 7c9b6379-e575-9819-eca1-a0952f66b655
* Version Independent ID: 34d81602-ffa2-5be6-67ef-b42b3b6f17ee
* Content: [Configure Microsoft Defender Antivirus exclusions on Windows Server 2016 or 2019 - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-antivirus/configure-server-exclusions-microsoft-defender-antivirus)
* Content Source: [windows/security/threat-protection/microsoft-defender-antivirus/configure-server-exclusions-microsoft-defender-antivirus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-antivirus/configure-server-exclusions-microsoft-defender-antivirus.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

illfated:
Good point. The "Applies to:" line is the only text referencing "Microsoft Defender Advanced Threat Protection (Microsoft Defender ATP)" within that document. The rest of the document seems to be purely Microsoft Defender Antivirus.

Oddly enough, I did not notice this last time I edited this document, but then again, I was mostly only focused on encapsulating system references in code blocks to avoid automatic translation in non-English pages.

e0i:
Closing via #8005

PR #8005, opened 2020-08-04
Removing of Applies to

joinimran:
As the applies to portion is linking to the commercial differences among Windows 10 and doesn't fit with the article so it has been removed to make document clear. 

Problem: https://github.com/MicrosoftDocs/windows-itpro-docs/issues/7106


joinimran:
@JohanFreelancer9 - kindly do the copy editor review for this PR. Thanks.

PR #8018, opened 2020-08-05
Add clarification for FAQs regarding Target Version policy

andreiztm:


andreiztm:
This PR has been discussed with @arcarley .

andreiztm:
Thank you for the fast action!

Issue #7134, opened 2020-07-10
Cloud resources


velden:
Unclear (I'd guess wrong) statement in the text:

"If you **don’t** use proxy servers, you must also **include** the "," delimiter just before the "|"."

Followed by an example without proxy NOT using the comma:
Value format without proxy:
`contoso.sharepoint.com|contoso.visualstudio.com`

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c1ad3f92-24de-5eb5-8454-c72d85737183
* Version Independent ID: 900125eb-31ea-faf3-ac30-7c3d7b30f611
* Content: [Create a Windows Information Protection (WIP) policy with MDM using the Azure portal for Microsoft Intune (Windows 10) - Microsoft 365 Security](https://docs.microsoft.com/en-us/windows/security/information-protection/windows-information-protection/create-wip-policy-using-intune-azure)
* Content Source: [windows/security/information-protection/windows-information-protection/create-wip-policy-using-intune-azure.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/information-protection/windows-information-protection/create-wip-policy-using-intune-azure.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @DulceMontemayor
* Microsoft Alias: **dansimp**

joinimran:
@Dansimp - Can you please share your input here. I couldn't get any exact information for the same.

@velden - If there is any information that you have and needs to be added, please let me know so that I can update the document.

Thanks.
Imran

e0i:
Fixed via #8087

PR #8087, opened 2020-08-18
Comma missing in example

joinimran:
As mentioned in the document that if we are adding a URL without proxy we need to add "," just before the "|". Added this in the documents example section. 

Problem: https://github.com/MicrosoftDocs/windows-itpro-docs/issues/7134


joinimran:
@JohanFreelancer9 - kindly do the copy editor review for this PR. Thanks.

Issue #8095, opened 2020-08-18
Incorrect query function


erroltuparker:
The query provided is no longer valid

New query is as per below

DeviceEvents
| where Timestamp > ago(7d) and
ActionType startswith "AppControl"
| summarize Machines=dcount(DeviceName) by ActionType
| order by Machines desc

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1a868903-95b1-288c-72d2-be936171044a
* Version Independent ID: 1456cb15-b4ec-2c0d-958e-340766c27788
* Content: [Query Application Control events with Advanced Hunting (Windows 10) - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-control/querying-application-control-events-centrally-using-advanced-hunting)
* Content Source: [windows/security/threat-protection/windows-defender-application-control/querying-application-control-events-centrally-using-advanced-hunting.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/windows-defender-application-control/querying-application-control-events-centrally-using-advanced-hunting.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jsuther1974
* Microsoft Alias: **dansimp**

illfated:
Ref. PR #8127

e0i:
Closing via #8127

PR #8127, opened 2020-08-24
Update WDAC hunting query

BenMcGarry:
Existing query does not appear to work within WDATP Advanced hunting, this updates the query to return the expected result.


illfated:
(ref. ticket #8095)

PR #8166, opened 2020-08-27
Add Update Baseline

katoma2017:
Adding a blurb and explanation of Update Baseline when it comes to helping improve end user experience


arcarley:
@jaimeo This has been reviewed and is ready to be added. 👍

PR #8168, opened 2020-08-27
Update commercial-gov.md

adirdidi:
Update the known gaps date.


mjcaparas:
Hi, would like to know why the required update for the date. 
I think it's valid to state 2019 because that covers items from that year onwards.

adirdidi:
I've refreshed the content of this page 2-3 times in the last month, so I think we should update the date.
(I got that a question from the field, as they think it's confusing)

PR #8183, opened 2020-08-31
Correcting -EncryptionMethod values

SteveBurkettNZ:
Correcting the -EncryptionMethod values to current accepted value (without the Xts prefix).


SteveBurkettNZ:
Hmmm...   Seems that XtsAes256 is valid, but getting a 'value does not fall within the expected range' error.  Aes256 works fine. Closing as I'm incorrect here sorry!

PR #8202, opened 2020-09-02
replaced mmd to mdm

RAJU2529:
as per the user report #8200  so I replaced mmd to mdm

Issue #8200, opened 2020-09-02
update needed to document


hagoldsh:
[Enter feedback here]
The page contains many links to Intune which should change to MEM (Intune no longer exist). 
also, there is a typo in the section title for Mobile Device Management. the acronyms in the parenthesis should be MDM and not MMD.

Thanks,
Haim

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 71868539-63d8-b250-f9d4-83a4052f4f74
* Version Independent ID: ed5658cf-173f-47bf-ed76-a68d89ea5c4b
* Content: [Turning on network protection - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/enable-network-protection?source=docs)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/enable-network-protection.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/enable-network-protection.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @levinec
* Microsoft Alias: **ellevin**

e0i:
The issue has been handled with a community contributed PR, #8204. Thank you for the discussion.

PR #8204, opened 2020-09-02
replaced mmd to mdm

RAJU2529:
as per user report #8200, so i replaced mmd to mdm,  this  PR is  different from #8202 ,


levinec:
Making this change in 8202 to keep the changes together.

PR #8249, opened 2020-09-09
Update manual config to include full census sync

cinglis-msft:
Incidents have been popping up wherein customers are experiencing issues with missing fields. This is partially due to Census not fully syncing those fields on every run, only once per week. The config script has always invoked a full Census sync. This is being added as it has been observed many customers simply manually configure devices for Update Compliance.


cinglis-msft:
@jaimeo adding additional bit of language to manual UC config to include directions on performing a full Census sync.

PR #8250, opened 2020-09-09
Remove product removal blurbs

cinglis-msft:
It's been 6 months since they were removed. @jaimeo


jaimeo:
You remembered! I sure didn't.

PR #8260, opened 2020-09-10
Update manage-updates-baselines-microsoft-defender-antivirus.md

kurtsarens:
Added a note where the customer can check the current versions.
Added a new item to the what's new list of 2008.9, updated version of platform


denisebmsft (approved):
Thanks @kurtsarens

PR #8281, opened 2020-09-14
Update waas-delivery-optimization-setup.md

WplusAzureAuto:
Amended line 149 from Get-DeliveryOptimizationPerfSnap to Get-DeliveryOptimizationStatus, this command is validated from windows powershell to be the valid command that has -peerinfo switch


WplusAzureAuto:
Amended line 149 from Get-DeliveryOptimizationPerfSnap to Get-DeliveryOptimizationStatus, this command is validated from windows powershell to be the valid command that has -peerinfo switch

illfated:
(ref. #5620)

PR #8286, opened 2020-09-15
MDATP for Android General Availability Sep 2020

shthota77:
This change needs to be pulished in Sep 21st.
Microsoft Defender for Android is announcing Generaly Availability on Sep 22nd 2020.


mjcaparas:
Will merge on sept 21.

PR #8287, opened 2020-09-15
MDATP Android GA: Remove Public Preview Note

shthota77:
MDATP Android GA is announced on 22nd Sep. This change needs to be published on Sep 21st evening US time.
Change: Deleted the Public Preview Note.


mjcaparas:
Will merge on sept 21.

PR #8291, opened 2020-09-15
Fix bold marking

Jonher937:
Simple fix. the "C" was not in bold.


illfated (approved):
Good catch, thanks. Must have been a marking error when the text was written originally.

illfated:
#sign-off

PR #8348, opened 2020-09-25
Updating MSA warning documentation

alecoot:
Removing top-level warning-banner for MSA as this has been fixed for a number of releases and only applies to Windows 10 versions 1703, 1709, and 1803.


arcarley:
@jaimeo could you commit these changes? Thanks!

Issue #8382, opened 2020-09-30
No ScanScheduleQuickTime option


KimRechnagel:
Hi,

There is no ScanScheduleQuickTime option for the Set-MpPreference cmdlet as stated in this article. I'm pretty sure the correct option you are referring to is called ScanScheduleQuickScanTime instead.

Kim

[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2aeeaa89-e3b9-7dd5-81f2-7cb1f37d925d
* Version Independent ID: 76f816fa-8acd-a260-7d4b-c1a541b785db
* Content: [Schedule regular quick and full scans with Microsoft Defender AV - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-antivirus/scheduled-catch-up-scans-microsoft-defender-antivirus)
* Content Source: [windows/security/threat-protection/microsoft-defender-antivirus/scheduled-catch-up-scans-microsoft-defender-antivirus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-antivirus/scheduled-catch-up-scans-microsoft-defender-antivirus.md)
* Product: **w10**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

RAJU2529:
@KimRechnagel . from my testing on windows 10 v2004  both options are available and valid.

KimRechnagel:
Yes that is true but find this section on the source page:
![image](https://user-images.githubusercontent.com/21287647/94672204-39920700-0315-11eb-80f5-b0b4a4551ed1.png)

Then you see that the page says ScanScheduleQuickTime and not ScanScheduleQuickScanTime

KimRechnagel:
Ah, nevermind, you already figured it out yourself. Thanks.

PR #8383, opened 2020-09-30
added the word  Scan in between  Quick and Time

RAJU2529:
as per the user report #8382 , so i added  the word  **Scan** in  between  Quick and Time.
i  confirmed after checking the commands in PowerShell 

here is the proof 
**https://user-images.githubusercontent.com/3296790/94670428-2a6c8280-0330-11eb-92dc-03cae5f3bd61.JPG**


denisebmsft (approved):
Thank you @RAJU2529

PR #8390, opened 2020-10-01
UC config script note, doesn't configure DO

cinglis-msft:
@jaimeo - we could also include a link to DO docs in the Note if you think that's best, not sure about the navpath.


jaimeo:
I will add a link to the Delivery Optimization info, but that'll be easier to do internally, since it will add the correct path automatically.

PR #8393, opened 2020-10-01
Update use-group-policy-microsoft-defender-antivirus.md

kurtsarens:
Allow users to pause scan is not supported on Windows 10


denisebmsft (approved):
Thanks @kurtsarens

PR #8405, opened 2020-10-02
clarifying behavior on network shares

martyav:
I'm an internal contributor. I had a request to update this page because there was customer confusion about what turning on network sharing does and doesn't do.

> In the Note:  If real-time protection is enabled, files are scanned before they are accessed and executed. The scanning scope includes all files, including those on mounted removable devices such as USB drives.
>
> https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-antivirus/configure-advanced-scan-types-microsoft-defender-antivirus
>
> We need to ensure we also mention that files on network shares will also be scanned when executed from a machine that has RTP/on access  enabled.  Customers are under the impression that with network scanning turned off, we will not scan anything on a network share ever.


e0i:

JohanFreelancer9 (approved):

Issue #6860, opened 2020-06-02
System/AllowLocation


icelava:
When I look at the corresponding sensors.admx, the DisableLocation_2 policy only has two options.

```
<policy name="DisableLocation_2" class="Machine" displayName="$(string.DisableLocation)" explainText="$(string.DisableLocation_Explain)" key="Software\Policies\Microsoft\Windows\LocationAndSensors" valueName="DisableLocation">
      <parentCategory ref="LocationAndSensors" />
      <supportedOn ref="windows:SUPPORTED_Windows7"/>
      <enabledValue>
        <decimal value="1" />
      </enabledValue>
      <disabledValue>
        <decimal value="0" />
      </disabledValue>
</policy>
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2c833e5b-0fcf-2ace-a90a-3270286d23a9
* Version Independent ID: 64afbc38-e7cc-fc14-9d57-1cd119b2395d
* Content: [Policy CSP - System - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-system)
* Content Source: [windows/client-management/mdm/policy-csp-system.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/policy-csp-system.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @ManikaDhiman
* Microsoft Alias: **dansimp**

MaratMussabekov:
Hello @icelava ,
the admx file will be changed, please see the response from Microsoft representative in #8406.
Thank you

PR #8406, opened 2020-10-03
changed GPO reference

MaratMussabekov:
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/6860

### DIFF
--- a/windows/client-management/mdm/policy-csp-system.md
+++ b/windows/client-management/mdm/policy-csp-system.md
@@ -581,10 +581,10 @@ For example, an app's original Location setting is Off. The administrator then s
 <!--/Description-->
 <!--ADMXMapped-->
 ADMX Info:  
--   GP English name: *Turn off location*
--   GP name: *DisableLocation_2*
--   GP path: *Windows Components/Location and Sensors*
--   GP ADMX file name: *Sensors.admx*
+-   GP English name: *Let Windows apps access location*
+-   GP name: *LetAppsAccessLocation*
+-   GP path: *Windows Components/App Privacy*
+-   GP ADMX file name: *Appprivacy.admx*
 
 <!--/ADMXMapped-->
 <!--SupportedValues-->


MaratMussabekov:
Dear @JohanFreelancer9,
kindly do the copy editor review for this PR.

Dear @nam31, for your information

Thank you

JohanFreelancer9 (approved):

nam31:
@ManikaDhiman PR has been copyedited and is ready for final review, could you please check and merge? Thanks!

Cheoljin:
@MaratMussabekov I am not aware of any recent changes associate with this file. Can you explain why we try to make the change? What is the expected gain by updating this? Do you think this was a bug?

MaratMussabekov:
Hello @Cheoljin ,
**Can you explain why we try to make the change?** - System/AllowLocation policy was referred to the wrong GPO. It duplicates functionality of *Let Windows apps access location* GPO, not of *Turn off location*
Thank you

Cheoljin:
@MaratMussabekov I am from the Microsoft feature team to review. From my view, we should not make this change. 

For context, there are two policies, similar each other but different: (1) System/AllowLocation (2) Policy/LetAppsAccessLocation.

(2) is newer and extended version, containing the feature of (1), in terms of global/device-wide location access control. But (2) has additional feature to control per-app basis, e.g., LetAppsAccessAccountInfo_UserInControlOfTheseApps_List.

Thus, AppPrivacy.admx and Sensors.admx have different purpose and definition; i.e., mapping values for (a) user-control. (b) force-on. (c). force-off are different; for example Privacy/LetAppsAccessLocation defines "User-in-control" as 0, while System/AllowLocation defines 1 as "User-in-control".

Interestingly, sensors.admx does not match this right definition, though. This only defines on and off and their values are not correct.

However, updating Sensors.admx to AppPrivacy.admx (assuming that they are duplicate) is not correct too, causing more confusion, e.g., customer set "User-in-control" but ends up being another, i.e., ForceOFF.

I'd find the right admx file to replace (that has the following definition: MDMForceOFF -> 0. MDMNoControl -> 1. MDMForceON -> 2). Unfortunately, I was not able to do within a reasonable time I spent. If we cannot find the right one, I'd keep this as is - do you have a good reason that we should update this this time?

MaratMussabekov:
Hello @Cheoljin , thank you for the detailed explanation.
I understand now why *Sensors.admx* shouldn't be replaced by *Appprivacy.admx* here. 
I proposed to change GP to LetAppsAccessLocation because I didn't know about Policy/LetAppsAccessLocation (as it is not presented in the original article).

However, the description of DisableLocation_2 GP is about the enabling/disabling the feature:
"If you enable this policy setting, the location feature is turned off, and all programs on this computer are prevented from using location information from the location feature.
If you disable or do not configure this policy setting, all programs on this computer will not be prevented from using location information from the location feature."
It is not related to the System/AllowLocation policy, which is about the possibility of the user to change the settings.  System/AllowLocation much more similar to the description of LetAppsAccessLocation GP:
"If you choose the "User is in control" option, employees in your organization can decide whether Windows apps can access location by using Settings > Privacy on the device.
If you choose the "Force Allow" option, Windows apps are allowed to access location and employees in your organization cannot change it.
If you choose the "Force Deny" option, Windows apps are not allowed to access location and employees in your organization cannot change it."

So if it is not possible to find the correct admx file right now, would it be possible to just remove "ADMX Info:" section for  System/AllowLocation policy? Or we cannot do this because of Policy CSP document template?

thank you

Cheoljin:
@MaratMussabekov Current description makes sense more to me (unfortunately). If you go to (Computer Configuration -> Administrative Templates -> Windows Components -> Location And Sensors -> Turn off location), you see that the GP UX only provides three options with the radio button; "Not configured", "Enabled", and "Disabled", which are corresponding to "User-in-control", "ForceOFF", and "ForceON".

With that context, the description you provided me about DisableLocation_2 GP makes sense to me (although it could be misleading). Note that GP name is "Disable"-Location*, indicating that it is asking if customer wants to 'forcefully' disable all the apps by controlling the higher-order status. Again, if it is not Enabled (i.e., either Not configured or Disabled), it is valid to indicate that user can change the location settings in the Location->Privacy->Location.

Thus, I would not change anything. Let me know your point of view, though.

MaratMussabekov:
hello @Cheoljin,
yes, sure, as you recommend, I can just close this PR, without any change.
Thank you

PR #8411, opened 2020-10-05
Update tvm-software-inventory.md

sbrondel:
I've worked with customers who are expecting a full SCCM-style Software Inventory of all clients because of the sentence "The software inventory in threat and vulnerability management is a list of all the software in your organization".  Edit adds on "with known vulnerabilities" to reflect the true scope of this inventory.


levinec:
This does make it clearer, thank you!

PR #8413, opened 2020-10-06
update networkservice table

narkissit:
NetworkService + netsh proxy - should be "yes" in the table instead of "no"


illfated (approved):
Looks good to me.

illfated:
#sign-off

PR #8441, opened 2020-10-11
Fix broken link

rgsteele:


illfated (approved):
Looks like a good solution to me. Link restoration is valid. Previous link is broken, likely due to directory renaming.

illfated:
#sign-off

Issue #8454, opened 2020-10-13
Wrong (old one) term of channel


MilanKrocek:
Hello, I see term "LTSB" it should be "LTSC" am I right?

Thanks.

MK

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ad58ab5a-035d-569f-5849-f29ee7eaf72b
* Version Independent ID: 74c7578e-fb9e-76f0-bfa7-81d51dfbebfb
* Content: [Prepare servicing strategy for Windows 10 updates (Windows 10) - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/update/waas-servicing-strategy-windows-10-updates)
* Content Source: [windows/deployment/update/waas-servicing-strategy-windows-10-updates.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/update/waas-servicing-strategy-windows-10-updates.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jaimeo
* Microsoft Alias: **jaimeo**

RAJU2529:
@MilanKrocek . Yes . What you said is  correct. Thanks for pointing out .
I am fixing the error.

illfated:
See PR #8456 for the initialism correction.

nam31:
Closing via merged PR #8456 . Thank you

PR #8456, opened 2020-10-13
renamed LTSB to LTSC

RAJU2529:
as per the user report #8454 ,so I renamed LTSB to LTSC


illfated (approved):
Looks correct to me. See also https://docs.microsoft.com/windows/whats-new/ltsc/

PR #8469, opened 2020-10-15
Case sensitivity, re-ordering and other small corrections

poortom1004:
- Normalized the casing from BuiltIn Local to Builtin Local for group types
- Corrected some other group types
- Corrected typo for references of Group Policy Creators Owners to Group Policy Creator Owners
- Re-ordered the Read-Only Domain Controllers group to be higher in the list to be correctly alphabetized so that it matches the order in the first table
- Corrected Guests group membership details
- Added missing SID info on a few groups
- Changed group types from Domain Global to Global
- Replaced "No" with "None" for default membership to be consistent with other groups
- RDS Endpoint Servers had an incorrect reference to the domain SID
- Users group incorrectly said it's a member of Domain users via primary group membership.  This is incorrect because groups do not have primary groups, only users have primary groups.

### DIFF
--- a/windows/security/identity-protection/access-control/active-directory-security-groups.md
+++ b/windows/security/identity-protection/access-control/active-directory-security-groups.md
@@ -1950,7 +1950,7 @@ This security group has not changed since Windows Server 2008.
 </tr>
 <tr class="even">
 <td><p>Type</p></td>
-<td><p>Domain local</p></td>
+<td><p>Domain Local</p></td>
 </tr>
 <tr class="odd">
 <td><p>Default container</p></td>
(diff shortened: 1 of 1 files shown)


e0i:

windows/security/identity-protection/access-control/active-directory-security-groups.md
```diff
@@ -1950,7 +1950,7 @@ This security group has not changed since Windows Server 2008.
 </tr>
 <tr class="even">
 <td><p>Type</p></td>
-<td><p>Builtin local</p></td>
+<td><p>Domain local</p></td>
```
JohanFreelancer9, line 1953:
```suggestion
<td><p>Domain Local</p></td>
```

JohanFreelancer9 (changes requested):
Copy review - one minor edit.

Issue #8474, opened 2020-10-16
Wrong supported value for "Update/AutomaticMaintenanceWakeUp"


ChizuruOkamoto:
Wrong supported value for "Update/AutomaticMaintenanceWakeUp"

The supported value is True or False, but isn't it a mistake of 0 or 1?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 91ca58bd-f279-7739-0140-c4bd56784768
* Version Independent ID: 1d2ecc90-58a3-6bcd-84ae-ef04d7a3b7b4
* Content: [Policy CSP - Update - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-update#update-automaticmaintenancewakeup)
* Content Source: [windows/client-management/mdm/policy-csp-update.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/policy-csp-update.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @ManikaDhiman
* Microsoft Alias: **dansimp**

RAJU2529:
@ChizuruOkamoto , Wrong values are entered,  what you said is correct. i am fixing it now.

RAJU2529:
proof 
![Uploading Capture.JPG…]()

nam31:
Closing via #8476 
Thanks

PR #8476, opened 2020-10-16
removed wrong values added correct values

RAJU2529:
as per the user report #8474 , so i  removed wrong values, added correct values.
i enabled this policy , and i verified on registry.

Capture

### DIFF
--- a/windows/client-management/mdm/policy-csp-update.md
+++ b/windows/client-management/mdm/policy-csp-update.md
@@ -1110,7 +1110,7 @@ ADMX Info:
 <!--/ADMXMapped-->
 <!--SupportedValues-->
 Supported values:  
--   0  - Disable
+-   0  - Disable (Default)
 -   1  - Enable
 <!--/SupportedValues-->
 <!--Example-->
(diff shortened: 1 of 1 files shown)


windows/client-management/mdm/policy-csp-update.md
```diff
@@ -1110,8 +1110,8 @@ ADMX Info:
 <!--/ADMXMapped-->
 <!--SupportedValues-->
 Supported values:  
--   true - Enable
--   false - Disable (Default)
+-   0  - Disable
```
illfated, line 1113: The disabled state is the default state, isn't it?
```suggestion
-   0  - Disable (Default)
```

illfated (approved):
Looks good to me.

e0i:

JohanFreelancer9 (approved):

ManikaDhiman:
Hi @Kateyanne, Thanks for merging this PR. I was actually awaiting confirmation from the policy owner as there is a discrepancy in the default supported value listed here and the DDF. I just got the confirmation that the update in this PR is correct, hence we are good with the merge. Thanks!

Kateyanne:
Hi @ManikaDhiman - Phew! I'm glad it worked out. Let me talk with @e0i about how this could have gone more smoothly for you, and to prevent in the future. Thanks for your patience.

PR #8498, opened 2020-10-20
Specify Expected Behavior for Username/Password Setting

angela-em:
As an Intune SE, we receive a lot of cases for this OMA-URI setting. Customers use it to create a local admin account for Intune enrolled devices but it always reports as failed.
By including this information, we can set the expectation and prevent future cases.

### DIFF
--- a/windows/client-management/mdm/accounts-csp.md
+++ b/windows/client-management/mdm/accounts-csp.md
@@ -52,6 +52,7 @@ This node specifies the username for a new local user account.  This setting can
 This node specifies the password for a new local user account.  This setting can be managed remotely. 
 
 Supported operation is Add.
+GET operation is not supported.  This setting will report as failed when deployed from the Endpoint Manager.
 
 <a href="" id="users-username-localusergroup"></a>**Users/_UserName_/LocalUserGroup**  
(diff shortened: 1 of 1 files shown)


e0i:
CC @JohanFreelancer9 for copy editing.

windows/client-management/mdm/accounts-csp.md
```diff
@@ -52,6 +52,7 @@ This node specifies the username for a new local user account.  This setting can
 This node specifies the password for a new local user account.  This setting can be managed remotely. 
 
 Supported operation is Add.
+GET operation is not supported.  This setting will report as failed when deployed from the Endpoint Manager.
```
JohanFreelancer9, line 55:
```suggestion
The Get operation is not supported. This setting will report as failed when deployed from the Endpoint Manager.
```

JohanFreelancer9 (changes requested):
Copy review.

illfated:
🤔 Oh well. I guess the readers will get (pun intended) the added message clearly enough. 😉

windows/client-management/mdm/accounts-csp.md
angela-em: How long does it take before the merged changes show in the public doc?

PR #8539, opened 2020-10-26
Update to include the 2004 'Direction' value

PeterSmithRedmond:
We added a Direction value (inbound and outbound) to help a bunch of customers


denisebmsft (approved):
Thank you @PeterSmithRedmond

PR #8540, opened 2020-10-26
Includes the body of the updated Windows, version 2004 DDF file

PeterSmithRedmond:
We've added several new features; adding them here. This is cut-n-pasted from the downloadable DDF file.


denisebmsft (approved):
Thank you @PeterSmithRedmond

PR #8563, opened 2020-11-02
Update commercial-gov.md

amirsc3:
Fixing typo to avoid customer confusion. Some customers have opened support tickets because they could not understand from the above phrase which "patch" is referred to in this note.


illfated:
Good correction & update. Well done. Thanks.

PR #8591, opened 2020-11-05
Update controlled-folders.md

kurtsarens:
boot sectors are also protected


denisebmsft (approved):
Thank you @kurtsarens

Issue #8609, opened 2020-11-09
Duplicate entry in "Windows 10 features we’re no longer developing"


hematogender:
There's a duplicate Microsoft Edge entry in [this](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/planning/windows-10-deprecated-features.md) document.

![image](https://user-images.githubusercontent.com/46577620/98599736-2a738100-22e5-11eb-85cd-8af020f7a242.png)

RAJU2529:
@hematogender . Which article are you talking about?

hematogender:
@RAJU2529 The one I linked -- https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/planning/windows-10-deprecated-features.md

RAJU2529:
@hematogender , Thanks for pointing out, I am fixing duplicate entry now.

illfated:
See PR #8613 for the request to remove the duplicate entry.

denisebmsft:
Thank you, all!

PR #8613, opened 2020-11-10
removed duplicate entry

RAJU2529:
as per the user report  #8609 , so I removed the duplicate entry which is appeared two times.


illfated (approved):
Looks good to me.

illfated:
#sign-off

Issue #8614, opened 2020-11-10
Examples of exploit kits hyperlinks are broken


jgeurten:
Examples of exploit kits:
Angler / Axpergle

Neutrino

Nuclear

Do not have valid URIs 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: cd212223-309f-aab6-c64f-440018592716
* Version Independent ID: b19dd624-8094-ac87-df90-d81cd517c150
* Content: [Exploits and exploit kits - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/intelligence/exploits-malware)
* Content Source: [windows/security/threat-protection/intelligence/exploits-malware.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/intelligence/exploits-malware.md)
* Product: **w10**
* GitHub Login: @levinec
* Microsoft Alias: **ellevin**

RAJU2529:
@jgeurten Thanks for pointing out. I am adding new links now.

levinec:
Edited and merged pull request. Thank you @jgeurten and @RAJU2529!

PR #8619, opened 2020-11-11
removed invalid links . added correct links

RAJU2529:
as per the user report #8614 , so i removed  three invalid links  and added correct links


levinec:
Thanks for opening this PR @RAJU2529! The main issue is actually that it needs a language tag (en-us) in the URL to work. It must have been removed. I'll add it back in.

Issue #8634, opened 2020-11-13
Broken links


SeeingBlue:
[Enter feedback here]

The following quote is listed on the below page. This quote contains two links which result in 404 responses.
> Learn how to use Pool Monitor and how to use the data to troubleshoot pool leaks.

The links are: https://support.microsoft.com/help/177415 & https://blogs.technet.com/b/markrussinovich/archive/2009/03/26/3211216.aspx

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b7aab622-1bde-637a-8999-cac5f3f68ac4
* Version Independent ID: 16af5e10-0748-ec6f-f539-2ee8563c3702
* Content: [Advanced troubleshooting for Windows-based computer freeze issues - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/troubleshoot-windows-freeze#use-pool-monitor-to-collect-data-for-the-physical-computer-that-is-no-longer-frozen)
* Content Source: [windows/client-management/troubleshoot-windows-freeze.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/troubleshoot-windows-freeze.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

illfated:
Closest I could get for the URL text "use the data to troubleshoot pool leaks":
- https://docs.microsoft.com/en-us/archive/blogs/markrussinovich/pushing-the-limits-of-windows-paged-and-nonpaged-pool#tracking-pool-leaks

illfated:
The support article is a dead end, at least as the page content and existence goes. WayBack Machine says "**Nothing found for this domain**".

nam31:
Closing via #8640 
Thanks

PR #8640, opened 2020-11-15
removed invalid link and added correct link

RAJU2529:
as per user report #8634 , so i removed invalid link and added the correct link.

### DIFF
--- a/windows/client-management/troubleshoot-windows-freeze.md
+++ b/windows/client-management/troubleshoot-windows-freeze.md
@@ -251,7 +251,7 @@ If the physical computer is still running in a frozen state, follow these steps 
 
 Pool Monitor shows you the number of allocations and outstanding bytes of allocation by type of pool and the tag that is passed into calls of ExAllocatePoolWithTag.   
 
-Learn [How to use Memory Pool Monitor to troubleshoot kernel mode memory leaks](https://support.microsoft.com/office/how-to-use-memory-pool-monitor-poolmon-exe-to-troubleshoot-kernel-mode-memory-leaks-4f4a05c2-ef8a-fca4-3ae0-670b940af398). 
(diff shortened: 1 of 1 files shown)


windows/client-management/troubleshoot-windows-freeze.md
```diff
@@ -251,7 +251,7 @@ If the physical computer is still running in a frozen state, follow these steps
 
 Pool Monitor shows you the number of allocations and outstanding bytes of allocation by type of pool and the tag that is passed into calls of ExAllocatePoolWithTag.   
 
-Learn [how to use Pool Monitor](https://support.microsoft.com/help/177415) and how to [use the data to troubleshoot pool leaks](https://blogs.technet.com/b/markrussinovich/archive/2009/03/26/3211216.aspx). 
+Learn [How to use Memory Pool Monitor to troubleshoot kernel mode memory leaks](https://support.microsoft.com/office/how-to-use-memory-pool-monitor-poolmon-exe-to-troubleshoot-kernel-mode-memory-leaks-4f4a05c2-ef8a-fca4-3ae0-670b940af398).
```
illfated, line 254: To me, it looks like there should either be a lowercase "how to" in the URL text, but the other option is to extend the "Learn " text to become a more descriptive phrase and point to the URL text with a colon. I would like to ask @JohanFreelancer9 for suggestions here, because I am not quite sure which solution is preferred here, to follow the MS Codestyle correctly.

illfated: Other than that, it looks like a good solution to provide a single link to one page covering more or less the same as both the previous links, especially given that the previous material is either outdated or gone.

JohanFreelancer9, line 254:
```suggestion
Learn [how to use Memory Pool Monitor to troubleshoot kernel mode memory leaks](https://support.microsoft.com/office/how-to-use-memory-pool-monitor-poolmon-exe-to-troubleshoot-kernel-mode-memory-leaks-4f4a05c2-ef8a-fca4-3ae0-670b940af398). 
```

JohanFreelancer9 (changes requested):
Copy review.

illfated (approved):
Looks like a useful and good solution to me.

JohanFreelancer9 (approved):

PR #8651, opened 2020-11-17
Adding more examples and Intune reporting bug

rogersoMS:
Adding changes. Awaiting peer review from Lindakup and ddsilva

### DIFF
--- a/windows/client-management/mdm/policy-csp-userrights.md
+++ b/windows/client-management/mdm/policy-csp-userrights.md
@@ -76,12 +76,27 @@ Here are examples of data fields. The encoded 0xF000 is the standard delimiter/s
   If you use Intune custom profiles to assign UserRights policies, you must use the CDATA tag (`<![CDATA[...]]>`) to wrap the data fields. You can specify one or more user groups within the CDATA tag by using 0xF000 as the delimiter/separator.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


windows/client-management/mdm/policy-csp-userrights.md
```diff
@@ -75,6 +75,9 @@ Here are examples of data fields. The encoded 0xF000 is the standard delimiter/s
    
   If you use Intune custom profiles to assign UserRights policies, you must use the CDATA tag (`<![CDATA[...]]>`) to wrap the data fields. You can specify one or more user groups within the CDATA tag by using 0xF000 as the delimiter/separator.
 
+> [!NOTE]
+> There is currently a reporting issue in the Microsoft Endpoint Manager (MEM) console which results in the setting reporting back a 'Remediation failed' (0x87d1fde8) error, even when the setting is successfully applied. To verify whether the setting has applied successfully, check the local Windows 10 device: Event Viewer>Applications and Services Logs<Microsoft>Windows>DeviceManagement-Enterprise-Diagnostics-Provider>Admin>Event ID 814. This issue is the result of the use of the CDATA tags, which are neccesary when more than a single entry is required. If there is only a single entry, the CDATA tags can be omitted - which will resolve the reporting false positive.
```
illfated, line 79:
```suggestion
> There is currently a reporting issue in the Microsoft Endpoint Manager (MEM) console which results in the setting reporting back a 'Remediation failed' (0x87d1fde8) error, even when the setting is successfully applied. To verify whether the setting has applied successfully, check the local Windows 10 device: Event Viewer **>** Applications and Services Logs \<Microsoft\> Windows **>** DeviceManagement-Enterprise-Diagnostics-Provider **>** Admin **>** Event ID 814. This issue is the result of the use of the CDATA tags, which are necessary when more than a single entry is required. If there is only a single entry, the CDATA tags can be omitted - which will resolve the reporting false positive.
```

Please note: Any word or text enclosed in Left Angle Bracket & Right Angle Bracket ( `< >` ) will be interpreted as an unknown HTML tag _and therefore not be displayed_. To avoid this, always precede this character combination with the escape character Backslash ( `\` ) to block this form of tag/code interpretation.

e0i:

illfated:
@Kateyanne : Thank you for the fast merge, although this file was not ready to merge yet. It will need re-editing, due to incorrect use of angle brackets ` < > ` (interpreted as code Markup Language tags when combined).

illfated:
(ref. #8707 )

PR #8678, opened 2020-11-20
Fixed typo

weijiechai:


illfated (approved):
Good catch, well spotted. Thank you.

PR #8688, opened 2020-11-22
Update windows-10-poc.md

weijiechai:
Fixed incorrect dash to hyphen.


illfated:
Oh. Right. So this change is about replacing the incorrect use of `<en dash>`
https://en.wikipedia.org/wiki/Dash#Common_dashes_and_their_Unicode_code-points
with the regular hyphen (https://en.wikipedia.org/wiki/Hyphen#Unicode).

illfated (approved):
While this change does not make much visual change, I presume this is important for copying commands and PowerShell cmdlets. I approve.

illfated:
#sign-off

PR #8689, opened 2020-11-22
Typo "**pass@word1**"→"**pass\@word1**"

hyoshioka0128:
Bold with escape characters

https://docs.microsoft.com/en-us/windows/deployment/windows-10-poc-sc-config-mgr

PR #8698, opened 2020-11-23
Typo "**pass@word1**"→"**pass\@word1**"

hyoshioka0128:
Bold with escape characters
related #8689

https://docs.microsoft.com/en-us/windows/deployment/deploy-windows-cm/deploy-windows-10-using-pxe-and-configuration-manager


illfated (approved):
Same issue as in #8689 and same solution, just a different file. It should not be glossed over that `@` is a special character in MarkDown formatting code interpretation.

illfated:
#sign-off

PR #8702, opened 2020-11-24
Update raw-data-export-storage.md

alons8:
Update the feature requirements.
Since this feature was moved to GA there is no need to enable preview features


illfated (approved):
Looks perfectly valid, correct and good to me.

PR #8707, opened 2020-11-25
Removing Intune/MEM specific details (not within scope of CSP docs)

rogersoMS:
Adding clarification on examples regarding Azure AD.
Removing Intune specific reporting issues as CSP documentation should be generic.


illfated:
(also ref. PR #8651 where the Note blob was added)

illfated (approved):
Looks correct and good to me.

JohanFreelancer9 (approved):

illfated:
#sign-off

PR #8736, opened 2020-12-01

cmknox:
Added support for Edge browser installations and updates.


illfated:
I don't know the specifics, but would an update like this warrant an update to the metadata date field as well?

cmknox:
thanks for asking, no need to update the metadata field for this change.

illfated (approved):
Looks good to me.

illfated:
#sign-off

PR #8771, opened 2020-12-07
adding images for auto-VPN

sunasing:


sunasing:
Will include these images once added in the repo.

sunasing:
#sign-off

PR #8806, opened 2020-12-14
Update edr-in-block-mode.md

BenAlfasi:
Fix


denisebmsft:
@BenAlfasi thank you for the edits! The EDR in block mode article looks great. I've requested a review from @mjcaparas because the other two articles are hers.

PR #8815, opened 2020-12-15
Update gov.md

adirdidi:
Add API section.


illfated:
@adirdidi : I notice you have used 3 backticks ( **```** ) before & after the links, to encapsulate them. Is there a MarkDown principle to use backticks this way? If so, I am not familiar with it and would like to know more about it. In my experience, 3 backticks are used on separate lines to create a code block for one or more lines/rows, like a code snippet or sample, or a whole script paragraph. Also, in my experience, it is only necessary to use 1 backtick before and after separate words or parts of a line, while the rest of the line (or row) stays unmodified by the backticks.

Hoping to hear from you about this, because this form of backtick usage has become a conundrum to me.

PR #8816, opened 2020-12-15
Update get-device-secure-score.md

adirdidi:
Add US Gov API note.


illfated:
Fair enough, good job. May I offer a word of advice regarding MarkDown?

It is generally recommended in MarkDown to add a blank space between the indent marker ( `>` ) and any following text. This is also known as "MarkDown indent marker compatibility spacing". I hope you may find the advice useful in your future use of the indent marker (also known as "quote marker").

PR #8832, opened 2020-12-17
Update mac-jamfpro-policies.md

cventour:
I have seen few customers using the example xml file as is , which will lead to exposing the device to risks (excluding pdfs , allowing users to create own exclusions , excluding the /home folder).

I modified the xml to include examples that would not expose the device to risk. Also added a comment for a customer to review the tags because if they dont change it all their devices will have the "ExampleTag" :)


illfated (approved):
Looks good to me.

illfated:
@cventour : Any particular reason you closed this? Are you planning on reopening, or maybe creating a new PR?

PR #8843, opened 2020-12-19
Move "lockdown" security info to the bottom

PeterSmithRedmond:
Per discussion with the VPN field team, lockdown is actually often a poor choice for customer -- it's more restrictive than many customer expect, leading to wasted time. It's still supported, but it should be at the bottom of the security list so that customer aren't instantly drawn to it.


illfated (approved):
Content move. New content is identical to old content. Looks good to me.

PR #8853, opened 2020-12-21
updated-4715491

Lovina-Saldanha:
updated the correct name "Microsoft Endpoint Manager"


illfated (approved):
7 corrections in 7 files (1 in 1): "Microsoft Endpoint Configuration Manager" (non-existing product) replaced with "Microsoft Endpoint Manager" (existing product).

Looks OK to me.

PR #8854, opened 2020-12-21

chriti:
We get a fair number of cases in support regarding Windows 10 Enterprise Subscription Based Activation and GCC/GCC High tenants. At this time it is not supported for GCC and GCC High Azure tenants. I would like this doc edited to reflect that so that we have a place to direct customers when this request comes up.

### DIFF
--- a/windows/deployment/windows-10-subscription-activation.md
+++ b/windows/deployment/windows-10-subscription-activation.md
@@ -229,6 +229,8 @@ Microsoft Products & Services Agreements (MPSA):
 - Existing MPSA customers will receive service activation emails that allow their customer administrator to assign users to the service.
 - New MPSA customers who purchase the Software Subscription Windows Enterprise E3 and E5 will be enabled for both the traditional key-based and new subscriptions activation method.
 
+** I would like to add a note here that this is only for comercial tenants and not GCC/GCC High tenants.**
+
 ### Deploying licenses
 
 See [Deploy Windows 10 Enterprise licenses](deploy-enterprise-licenses.md).


windows/deployment/windows-10-subscription-activation.md
```diff
@@ -229,6 +229,8 @@ Microsoft Products & Services Agreements (MPSA):
 - Existing MPSA customers will receive service activation emails that allow their customer administrator to assign users to the service.
 - New MPSA customers who purchase the Software Subscription Windows Enterprise E3 and E5 will be enabled for both the traditional key-based and new subscriptions activation method.
 
+** I would like to add a note here that this is only for comercial tenants and not GCC/GCC High tenants.**
+
```
illfated, line 233: OK, I see. This calls for an approved text from the team management, I presume.

chriti:
Yes it does. We have been getting support cases on it but have no documentation to send customers to when they want it. So we should get this doc or another doc updated to reflect that its not supported on GCC/GCC High tenants.



Thanks,
Chris Tiengo
Support Escalation Engineer | Support Engineering -  Cloud & Infrastructure Solutions | Devices Team
Customer Services & Support | Microsoft Corporation
* (425) 722-2100<tel:4257222100> | * chriti@microsoft.com<mailto:chriti@microsoft.com> | Teams/Skype IM<sip:chriti@microsoft.com>
Working Hours:Sunday - Wednesday 9AM - 8PM EST

You may also call (800) 936-3100 and ask for the next available support engineer. Be sure to have your SR number available when you call to ensure they are able to reroute your case to the queue for assignment.
If you have any concerns or feedback, you may reach my manager:  Marcie Davis |* marcied@microsoft.com<mailto:marcied@microsoft.com> |* +1 (980) 776-7524

illfated:
Fair point. I would have suggested something if it had not been so important in relation to real customer scenarios, but this is too important for me to come up with a reasonable suggestion. The only extra information I can add is that a follow-up might be slightly delayed due to the holidays, but I don't know (I'm an external contributor, not affiliated with MS).

greg-lindsay:
Hi, we need to check this with the product team since it is a direct statement about support. Also, we do not use first person like this (we don't use "I would like to..." because who is "I" in an article on docs?  Sometimes we specify that "Microsoft supports" or doesn't support a certain scenario.

greg-lindsay:
I am going to close this and we can work to add information to the article. Please contact me directly.

illfated:
Just to add my own view (although I understand the need to elaborate this in the background): The intention was for the MS Docs team to suggest a suitable phrase to replace the entered text in that note.

PR #8896, opened 2021-01-01
Fix typo MTD -> MDT

PeterUpfold:


illfated (approved):
Good typo correction. Thank you.

(for those who might have forgotten: MDT = Microsoft Deployment Kit)

PR #8944, opened 2021-01-11
Updated Applies to text

eross-msft:
Updated supported operating systems to include the ones from the previous file we redirected due to formatting issues.

### DIFF
--- a/windows/security/information-protection/bitlocker/bitlocker-group-policy-settings.md
+++ b/windows/security/information-protection/bitlocker/bitlocker-group-policy-settings.md
@@ -20,7 +20,7 @@ ms.custom: bitlocker
 
 # BitLocker Group Policy settings
 
-**Applies to**
+**Applies to:**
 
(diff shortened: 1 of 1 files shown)


windows/security/information-protection/bitlocker/bitlocker-group-policy-settings.md
```diff
@@ -22,7 +22,7 @@ ms.custom: bitlocker
 
 **Applies to**
```
illfated, line 23: Could we have a colon here while we are here in this section, or do you find that to be redundant? (I'm not 100% sure myself.)
```suggestion
**Applies to:**
```

eross-msft: I actually missed that it wasn't there. Thanks for the catch!

illfated (approved):
Thank you for the addition. Looks good to me. Nothing more from my side.

PR #8945, opened 2021-01-11
Title text change to new naming

wattbt:
The web page title was left still referring to Defender ATP, so updated for Defender for Endpoint. Not sure if other pages also are missing this change when the rest of the page was updated.


illfated:
Thank you for the contribution. I don't wish to ruin your day or something bad like that, but there is a lot more where that came from:
- https://github.com/MicrosoftDocs/windows-itpro-docs/search?q=%22Defender+ATP%22

**406 code results in [MicrosoftDocs/windows-itpro-docs](https://github.com/MicrosoftDocs/windows-itpro-docs)** or view [all results on GitHub](https://github.com/search?q=%22Defender+ATP%22)

Dansimp:
We are definitely working on rebranding across topics. We made a rather large push last few weeks, but we might have missed some. We are holding off on changing filenames/folders until we get a larger set of redirections set up so we don't break links everywhere. Coming soon.  Thank you!

PR #8958, opened 2021-01-13
Update live-response-command-examples.md

formmailer:
Missing backslash in the command to view a registry value.

### DIFF
--- a/windows/security/threat-protection/microsoft-defender-atp/live-response-command-examples.md
+++ b/windows/security/threat-protection/microsoft-defender-atp/live-response-command-examples.md
@@ -158,7 +158,7 @@ registry HKEY_CURRENT_USER\Console
 
 ```
 # Show information about a specific registry value
-registry HKEY_CURRENT_USER\Console\ScreenBufferSize
+registry HKEY_CURRENT_USER\Console\\ScreenBufferSize
 ```
 
 


illfated:
This is from the document page
- https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/live-response-command-examples

More specifically, it is in the following paragraph:
- https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/live-response-command-examples#registry .

The relevant code samples in that page currently look like the following:

image

By adding the extra backslash, it looks like there will be a visible double backslash before `ScreenBufferSize` in the second "Copy" command code view, but I guess you are talking about how that command works when copied and applied to the OS in question, to produce the required result. Is this the case?

formmailer:
Hi,

No, the extra backslash should be visable, since it is required to retrieve a registry value. The single backslash is to retrieve a registry key. 
I attach a screen shot from the help command:
image

Best regards,
Jasper

illfated:
Very well, thank you for the useful reply. I see what you mean.

illfated (approved):
Looks correct to me, given the descriptions in this PR discussion.

illfated:
#sign-off

PR #8964, opened 2021-01-14
Update alerts.md

athaynes:
Update alert and return type links were incorrect.


illfated (approved):
Good correction, well spotted. Thank you.

PR #8970, opened 2021-01-14
Update web-content-filtering.md

msft-cjeich:
Text update to account for changes to our data provider.  
Also - 15 min for policy sync is incorrect.  SLA should be 2hrs. 
Include other 3rd party browsers which NP supports.  This is a server-side change which requires no client side changes.

### DIFF
--- a/windows/security/threat-protection/microsoft-defender-atp/web-content-filtering.md
+++ b/windows/security/threat-protection/microsoft-defender-atp/web-content-filtering.md
@@ -54,7 +54,7 @@ Before trying out this feature, make sure you have the following requirements:
 - Access to Microsoft Defender Security Center portal
 - Devices running Windows 10 Anniversary Update (version 1607) or later with the latest MoCAMP update.
 
-If Windows Defender SmartScreen isn't turned on, Network Protection will take over the blocking. It requires [enabling Network Protection](enable-network-protection.md) on the device.  Chrome, Firefox, Brave and Opera are currently 3rd party browsers in which the feature is enabled. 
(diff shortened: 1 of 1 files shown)


windows/security/threat-protection/microsoft-defender-atp/web-content-filtering.md
```diff
@@ -54,11 +54,11 @@ Before trying out this feature, make sure you have the following requirements:
 - Access to Microsoft Defender Security Center portal
 - Devices running Windows 10 Anniversary Update (version 1607) or later with the latest MoCAMP update.
 
-If Windows Defender SmartScreen isn't turned on, Network Protection will take over the blocking. It requires [enabling Network Protection](enable-network-protection.md) on the device.
+If Windows Defender SmartScreen isn't turned on, Network Protection will take over the blocking. It requires [enabling Network Protection](enable-network-protection.md) on the device.  Chrome, Firefox, Brave and Opera are currently 3rd party browsers in which the feature is enabled.
```
illfated, line 57: Consistency/specificity: I consider "this" to be a more specific and "to the point" reference than "the". Maybe I read documentation differently than most. May I offer this suggested change?
```suggestion
If Windows Defender SmartScreen isn't turned on, Network Protection will take over the blocking. It requires [enabling Network Protection](enable-network-protection.md) on the device. Chrome, Firefox, Brave, and Opera are currently 3rd party browsers in which this feature is enabled.
```
(Also removed some redundant spacing and added an Oxford comma.)

PR #8978, opened 2021-01-18
Update exposed-apis-odata-samples.md

SujudAbuAtta:
Added an example of Get devices with computerDnsName starting with


e0i:
Hello, @SujudAbuAtta Thank you for the contribution. The CLA needs to be signed before continuing with the review of this PR.

illfated (approved):
Looks quite OK to me, format-wise. I don't know the content well enough to verify its validity, but it seems to follow existing codestyle.

PR #9005, opened 2021-01-21

cmknox:
Add Edge browser support to content type table.

### DIFF
--- a/windows/deployment/update/waas-delivery-optimization.md
+++ b/windows/deployment/update/waas-delivery-optimization.md
@@ -66,7 +66,7 @@ For information about setting up Delivery Optimization, including tips for the b
     - Xbox game pass games
     - MSIX apps (HTTP downloads only)
     - Edge browser installs and updates
-    - Dynamic updates
+    - [Dynamic updates](https://techcommunity.microsoft.com/t5/windows-it-pro-blog/the-benefits-of-windows-10-dynamic-update/ba-p/467847) 
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


windows/deployment/update/waas-delivery-optimization.md
```diff
@@ -65,7 +65,7 @@ For information about setting up Delivery Optimization, including tips for the b
     - Office installations and updates
     - Xbox game pass games
     - MSIX apps (HTTP downloads only)
-    - Edge browser installations and updates
+    - Edge browser installs and updates
```
illfated, line 68: What is the intention behind changing "installations" to "installs"? Is it a different semantic (meaning) from the existing wording?

cmknox: Nope, just a preference.  Do you have an opinion?  I could go either way, but prefer `installs`.

illfated: Well, at first I was going to say that "installs" reads more often as a verb in the present tense than a noun, but I guess that is more the case for books and conservative writing. From what I have understood during the last year, it is the intention to keep the documentation less formal than traditional conservative books.

As long as "installs" is easily understood to be a noun in this case, it should be OK. I might not know the exact interpretation for some phrases, given that English is not my first language (although I read & write it on a near-native level).

cmknox: Ok, let's go with it and change it later, if needed.

illfated (approved):
Looks quite OK to me.

#sign-off

illfated (approved):
Looks like a good addition to have the added content linked to its relevant blog post.

PR #9016, opened 2021-01-25
Update create-wip-policy-using-intune-azure.md

jeffgilb:
Updated per request from PM (dereka).


illfated (approved):
Looks valid & good to me. Simple addition of a missing pipe character separator in the command.

PR #9021, opened 2021-01-26
fix title

peterlewis:
fix title which omitted full wording (replace "Set up Microsoft c for macOS device groups in Jamf Pro" with "Set up Microsoft Defender for Endpoint for macOS device groups in Jamf Pro")


illfated (approved):
Looks like a good correction to me. Also looks like the old "c" character could have been the result of a hastened copy & paste accident, but I am definitely glad that it gets corrected.

illfated:
#sign-off

denisebmsft (approved):
Thanks!

PR #9052, opened 2021-01-31
Added Conditional Access with Intune

sunasing:
Added details about Conditional Access with Intune
Re-arranged the sections.


sunasing:
#sign-off

illfated:
I see that this is mostly about rearranging some of the content, but it would still be useful to clarify a couple of details: I seem to remember that there is a preference for using "select" instead of "click or tap", but I would like to know if I am remembering this correctly before I suggest any changes.

sunasing:
This is a mobile application so "tap" is the ideal action that user does. Having said that I am fine with the guidelines that you mentioned. Let me know which one is more appropriate and I will change it.

illfated:
Thank you for replying. Seems that this has been handled internally, so I only managed to return to this PR after it has been merged. Fair enough. Thank you for the PR. 👍🏻

PR #9071, opened 2021-02-03
Update controlled-folders.md

amirsc3:
Some customers opened support tickets wanting to know why CFA blocks did not create alerts in our portal... so I think we should add this note to avoid customer confusion...


denisebmsft (approved):
Thank you @amirsc3

illfated:
Nice update. Thank you! 👍🏻

beerisgood:
Isn't 
> Protect important folders **against** ransomware 

Instead of
> Protect important folders from ransomware 

Better?

illfated:
Sure, it is slightly better and more precise from a grammar perspective. Then again, it is almost on a nuance level.

PR #9083, opened 2021-02-05
Update /remoteGuard requirements

piaudonn:
/remtoteGuard does not require to be a member of the local admintrators group of the source computer nor the target computer. There used to be a bug which was corrected (discussed with Andrew Wiley from Enterprise & Security SWE R&D)
However, it does require the user to have some RDP permissions at the remote computer level. So I updated the note to reflect the real requirements.

### DIFF
--- a/windows/security/identity-protection/remote-credential-guard.md
+++ b/windows/security/identity-protection/remote-credential-guard.md
@@ -181,7 +181,7 @@ mstsc.exe /remoteGuard
 ```
 
 > [!NOTE]
-> The user must be part of administrators group.
+> The user must be authorized to connect to the remote server using Remote Desktop Protocol. For example, by being a member of the Remote Desktop Users local group on the remote computer.
 
 ## Considerations when using Windows Defender Remote Credential Guard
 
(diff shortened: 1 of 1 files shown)


windows/security/identity-protection/remote-credential-guard.md
```diff
@@ -181,7 +181,7 @@ mstsc.exe /remoteGuard
 ```
 
 > [!NOTE]
-> The user must be part of administrators group.
+> The user must be authorized to connect to the remote server using Remote Desktop Protocol. For example, by being a member of the Remote Desktop Users local group on the remote computer.
```
illfated, line 184: The content in itself is good. From a grammatical and readability perspective, though, it feels like the second sentence is a fragment when reading it. Suggestion:
```suggestion
> The user must be authorized to connect to the remote server using Remote Desktop Protocol, for example by being a member of the Remote Desktop Users local group on the remote computer.
```
(@JohanFreelancer9 could have an even better suggestion for this line, if/when available)

piaudonn: Looks good :) Thanks!

illfated (approved):
Looks good to me.

JohanFreelancer9 (approved):

illfated:
#sign-off

PR #9090, opened 2021-02-05
Windows 10 Release Page url is changed after Base URL and redirection…

DocsPreview:
… change so we need to update all the existing urls to new urls.


DocsPreview:
Please do not merge these changes until 2/8 monday morning after 10:15 am only.

PR #9112, opened 2021-02-09
Changes to ASR licensing requirements

anvascon:
ASR rules don't require an E5 license. It's recommended given the extras like reporting and hunting.


denisebmsft (approved):
Thank you @anvascon

Issue #9111, opened 2021-02-09
Incomplete step listed explaining how to reach exclusions in virus & threat protection settings


maskedmustelid:
In https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-antivirus/microsoft-defender-security-center-antivirus under 'Add exclusions for Microsoft Defender Antivirus in the Windows Security app'

Step three is incompletely listed and just a bit too vague, easily leading to a failure to find the exclusion settings due to the instructions not being specific enough.

3 - Click Virus & threat protection settings.

Should be along the lines of:

3 - Click 'Manage settings' under Virus & threat protection settings subheading.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 8141d691-eee8-8861-4d04-24a78c7636b0
* Version Independent ID: 47c49eef-0812-2224-e8c4-fa4385df0789
* Content: [Microsoft Defender Antivirus in the Windows Security app - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-antivirus/microsoft-defender-security-center-antivirus)
* Content Source: [windows/security/threat-protection/microsoft-defender-antivirus/microsoft-defender-security-center-antivirus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-antivirus/microsoft-defender-security-center-antivirus.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

RAJU2529:
@maskedmustelid . Thanks for pointing out, I am fixing now .

e0i:
Thank you for your ongoing contributions, @RAJU2529  Much appreciated.

PR #9124, opened 2021-02-11
added  Manage Settings

RAJU2529:
as per the user feedback, #9111 , so i added  **manage settings**


denisebmsft (approved):
Thank you @RAJU2529

Issue #9104, opened 2021-02-08
Windows Update - Dead links


Shwaffle:
Under _How do I reset Windows Update components?_

The script links are dead.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 886426b5-8548-1cfc-ccd2-fd03e549c967
* Version Independent ID: 599aea76-47e4-dccc-799c-2d8acf549546
* Content: [Windows Update - Additional resources - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/update/windows-update-resources)
* Content Source: [windows/deployment/update/windows-update-resources.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/update/windows-update-resources.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jaimeo
* Microsoft Alias: **jaimeo**

illfated:
From https://docs.microsoft.com/windows/deployment/update/windows-update-resources#how-do-i-reset-windows-update-components

## How do I reset Windows Update components?

[Reset Windows Update Client settings script](https://gallery.technet.microsoft.com/scriptcenter/Reset-WindowsUpdateps1-e0c5eb78) will completely reset the Windows Update client settings. It has been tested on Windows 7, 8, 10, and Windows Server 2012 R2. It will configure the services and registry keys related to Windows Update for default settings. It will also clean up files related to Windows Update, in addition to BITS related data.

[Reset Windows Update Agent script](https://gallery.technet.microsoft.com/scriptcenter/Reset-Windows-Update-Agent-d824badc) allows you to reset the Windows Update Agent, resolving issues with Windows Update.

***

Note that the script links point to gallery.technet.microsoft.com — a retired site address now redirecting to https://docs.microsoft.com/samples/browse/?redirectedfrom=TechNet-Gallery (**Browse code samples**).

I seem to recall that at least one of the 2 scripts is present on GitHub in an updated form, but I am not sure if that is the solution we want to use for the docs.microsoft.com pages. It would have been better to link the original scripts, if they can be found in the archive pages.

illfated:
From the WayBack Machine / Web Archives:
1. https://web.archive.org/web/20201204152233/https://gallery.technet.microsoft.com/scriptcenter/Reset-Windows-Update-Agent-d824badc (Reset Windows Update Client settings script)
[...]

PR #9125, opened 2021-02-11
added new links

RAJU2529:
as per user report #9104 , so i added two new links.
i say thanks to @illfated  for given  links hint

### DIFF
--- a/windows/deployment/update/windows-update-resources.md
+++ b/windows/deployment/update/windows-update-resources.md
@@ -32,9 +32,9 @@ The following resources provide additional information about using Windows Updat
 
 ## How do I reset Windows Update components?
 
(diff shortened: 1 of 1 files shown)


illfated:
Wish it would be as easy as that, but web.archive.org does not belong to Microsoft nor any subsidiary of Microsoft. Therefore, those links are less likely to be accepted. ( I don't know yet, not 100% sure.) If you could find the GitHub.com pages, within the MicrosoftDocs repository, those links would be more likely to become accepted.

RAJU2529:
@illfated , Thanks  for the reply, we will see what will be the final decision from the editing team.

jaimeo:
No, we cannot link to sites not controlled by Microsoft. I have queries out to what is supposed to be the substitute for the retired Script Center. We can move the scripts there.

Issue #9135, opened 2021-02-12
Typo in article https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/configure-machines-security-baseline


darolt1:
There is a typo with this:
365 Microsoft 365 Defender

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d2246e9c-193b-a110-ca52-2110a6e63631
* Version Independent ID: 7328d4a4-1944-c199-db72-a7e3cd350ae9
* Content: [Increase compliance to the Microsoft Defender ATP security baseline - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/configure-machines-security-baseline)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/configure-machines-security-baseline.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/configure-machines-security-baseline.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @lomayor
* Microsoft Alias: **lomayor**

RAJU2529:
@darolt1 . Thanks for pointing out. I am fixing  now.

RAJU2529:
@e0i . I could not able to edit in content source site,  but I editted in other site . , luckily I was successfully editted.

darolt1:
Thanks!

denisebmsft:
This is fixed, and the updates will show up on the live site later today. Thank you so much for letting us know about this!

darolt1:
Great - thanks!

PR #9136, opened 2021-02-12
typo correction removed  365

RAJU2529:
as per user report #9135 , so i removed  365


denisebmsft (approved):
Thank you @RAJU2529. Just for clarification, we don't have Defender for Office, but rather, Defender for Office 365. I made that fix. Appreciate your help with the content!

PR #9187, opened 2021-02-22
Update hello-hybrid-aadj-sso-cert.md

aseigler:
IIS_IUSRS requires SeImpersonatePrivilege or IIS will throw a 500 error in AUTHENTICATE_REQUEST with "Either a required impersonation level was not provided, or the provided impersonation level is invalid. (0x80070542)"


illfated (approved):
Looks like a good solution to me, given the facts about the error scenario.

View the **Rich Diff** for easier review of the added content:
- https://github.com/MicrosoftDocs/windows-itpro-docs/pull/9187/files?short_path=d9ed6f9#diff-d9ed6f90ddd45e75edea4e012d3cffd794a23be08576e24355548a1f1e879c16

illfated:
@JohanFreelancer9 : Would you be available to perform a copy review on this PR content and its neighboring lines?

JohanFreelancer9 (approved):

e0i:
@mapalko Any comments on this PR?

mapalko:
I am not aware of needing this step. As far as I am aware manually adding this is typically not needed and I am hesitant about including this as general guidance.

aseigler:
It's absolutely required if you are using a service account to run the app pool that is not already otherwise privileged on the system, otherwise the user impersonation fails.

> I am not aware of needing this step. As far as I am aware manually adding this is typically not needed and I am hesitant about including this as general guidance.

mapalko:
I've never had to do this when doing test deployments. I checked with some customers engineers and they mentioned the same.

Instead of always listing it as a step, can we add it as a note with the context you mentioned? "if you are using a service account to run the app pool that is not already otherwise privileged on the system, otherwise the user impersonation fails."

PR #9201, opened 2021-02-24
Update microsoft-defender-antivirus-compatibility.md

kurtsarens:
EDR Block mode is a feature not a state... having it in the columns confuses people and creates support calls


denisebmsft (approved):
Thank you @kurtsarens

PR #9203, opened 2021-02-24
Add .circleci/config.yml

Dmarch28:


JohanFreelancer9 (commented):
Will do copy edit, if required, after technical review.

denisebmsft:
Not sure what this submission is for.

illfated:
This type of configuration is mostly used in repositories where one or more executable files are built as a result of compiling a dedicated source build.

Since MicrosoftDocs is not about compiling, I don't see a need for this type of config file, unless it can be documented that there are PowerShell files in need of testing before publishing to the docs.microsoft.com site.

@Dmarch28 : Do you have any documentation to show us how this can improve anything at all in the MS Docs repository?

dstrome:
Closing PR. We don't accept changes of this type.

Issue #9075, opened 2021-02-04
Reset Windows update components says to run CMD, but not in Admin mode


Susko3:
The first step says to open `cmd`, but doesn't specify it should be opened as an administrator. Most, if not all of the commands on that page require running as an administrator. This can be done by pressing Ctrl+Shift+Enter on the Run prompt, or in the Win+X menu (PowerShell Admin).

---
#### Document details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 886426b5-8548-1cfc-ccd2-fd03e549c967
* Version Independent ID: 599aea76-47e4-dccc-799c-2d8acf549546
* Content: [Windows Update - Additional resources - Windows Deployment](https://docs.microsoft.com/en-gb/windows/deployment/update/windows-update-resources)
* Content Source: [windows/deployment/update/windows-update-resources.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/update/windows-update-resources.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jaimeo
* Microsoft Alias: **jaimeo**

PR #9248, opened 2021-03-04
open Cmd as Administrator

VLG17:
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/9075


illfated (approved):
Good point. I have forgotten this line too often due to the discussions over what to do about the gallery.technet links, but that will be handled elsewhere. Anyway, thanks for the PR to resolve this line.

Issue #9109, opened 2021-02-09
Create and manage device groups - API


xenolf:
Hello!

I would like to enquire about the link to [Get list of tenant device groups using Graph API](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/get-machinegroups-collection) in the related topics section.

This link does not seem to lead anywhere and device group creation / querying is not mentioned anywhere else in the docs. Is this removed or preview functionality?

Thank you!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c912a9f9-2198-07ff-3ecb-e4d025b7c958
* Version Independent ID: 2781f02e-3869-498d-69fa-fc73d01ff661
* Content: [Create and manage device groups in Microsoft Defender ATP - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/machine-groups)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/machine-groups.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/machine-groups.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @mjcaparas
* Microsoft Alias: **macapara**

VLG17:
@xenolf It should point to [this article](https://docs.microsoft.com/en-us/graph/api/device-list-memberof?view=graph-rest-1.0&tabs=http). I'll create a change request for this.

PR #9249, opened 2021-03-04
update link

VLG17:
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/9109


illfated:
Could you check a couple of your browser versions and see if it still works with this shortened version of the URL?
- https://docs.microsoft.com/graph/api/device-list-memberof

This link works OK in my main browser, removing the /en-us/ language-specific part as well as the added tail `?view=graph-rest-1.0&tabs=http` (which I guess could be a page view CSS adaption).

VLG17 (commented):
as recommended by illfated

PR #9255, opened 2021-03-05
Update vpn-connection-type.md

PeterSmithRedmond:
Help fix some customer confusion -- the adapter is called "SSTP" even though it's also used for the plug-in protocols. We have a customer that doesn't want to use a particular VPN plug-in because of concerns that it's using or attempting to use the wrong protocol.

### DIFF
--- a/windows/security/identity-protection/vpn/vpn-connection-type.md
+++ b/windows/security/identity-protection/vpn/vpn-connection-type.md
@@ -41,7 +41,10 @@ There are many options for VPN clients. In Windows 10, the built-in plug-in and 
 
     - [SSTP](https://technet.microsoft.com/library/ff687819.aspx)
 
-        SSTP is supported for Windows desktop editions only. SSTP cannot be configured using mobile device management (MDM), but it is one of the protocols attempted in the **Automatic** option. Note: when a VPN plug-in is used, the adapter will be listed as an SSTP adapter even though the VPN protocol used is the plug-in's protocol.
+        SSTP is supported for Windows desktop editions only. SSTP cannot be configured using mobile device management (MDM), but it is one of the protocols attempted in the **Automatic** option.
+
+        > [!NOTE]
+        > When a VPN plug-in is used, the adapter will be listed as an SSTP adapter, even though the VPN protocol used is the plug-in's protocol.
         
 - Automatic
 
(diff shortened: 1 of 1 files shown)


windows/security/identity-protection/vpn/vpn-connection-type.md
```diff
@@ -41,7 +41,7 @@ There are many options for VPN clients. In Windows 10, the built-in plug-in and
 
     - [SSTP](https://technet.microsoft.com/library/ff687819.aspx)
 
-        SSTP is supported for Windows desktop editions only. SSTP cannot be configured using mobile device management (MDM), but it is one of the protocols attempted in the **Automatic** option.
+        SSTP is supported for Windows desktop editions only. SSTP cannot be configured using mobile device management (MDM), but it is one of the protocols attempted in the **Automatic** option. Note: when a VPN plug-in is used, the adapter will be listed as an SSTP adapter even though the VPN protocol used is the plug-in's protocol.
```
illfated, line 44: Based on previous experience with MS Docs editing, I am fairly certain that Notes should be added in form of colored Note blobs, like this:
```suggestion
        SSTP is supported for Windows desktop editions only. SSTP cannot be configured using mobile device management (MDM), but it is one of the protocols attempted in the **Automatic** option.

        > [!NOTE]
        > When a VPN plug-in is used, the adapter will be listed as an SSTP adapter, even though the VPN protocol used is the plug-in's protocol.
```

JohanFreelancer9 (changes requested):
Only @illfated's suggestion.

msarcletti:
Hi Peter,

you made the Note in the section for the 
Built-in VPN client - https://docs.microsoft.com/en-us/windows/security/identity-protection/vpn/vpn-connection-type#built-in-vpn-client 

I think it would better be located below
Universal Windows Platform VPN plug-in - https://docs.microsoft.com/en-us/windows/security/identity-protection/vpn/vpn-connection-type#universal-windows-platform-vpn-plug-in 
because it specifically targets third party VPN plug-in's.

And I suggest adding the same note in the FAQ section here:
Transition to OpenVPN protocol or IKEv2 from SSTP - https://docs.microsoft.com/en-us/azure/vpn-gateway/ikev2-openvpn-from-sstp#frequently-asked-questions

Regards
Markus

Kateyanne:
Peter says: I’m happy with it.

PR #9298, opened 2021-03-12
Update safety-scanner-download.md

nourdinb:
MSERT is supported for Windows Server 2019, isn't it?
That's what I have added in the "System requirements" section.
If it's not supported then we should state that as well.


JohanFreelancer9 (approved):

levinec:
Waiting for PM confirmation.

illfated (approved):

PR #9306, opened 2021-03-15
Update onboard-windows-10-multi-session-device.md

Bernd-Loehlein:
Changed old name to Defender ATP as Microsoft Defender for Endpoint was stated as old and new name.


adirdidi:
Thank you, that note should be removed completely (already doing that).

Issue #9333, opened 2021-03-19
i am going to edit


RAJU2529:
after reading this article, I want to afd links , making bold sentences. 
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f25443de-57f4-c8f2-fecb-f58714467894
* Version Independent ID: 1d783d00-ae0d-3eea-9a32-b42dd32e73c2
* Content: [Use the command line to manage Microsoft Defender Antivirus - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-antivirus/command-line-arguments-microsoft-defender-antivirus)
* Content Source: [windows/security/threat-protection/microsoft-defender-antivirus/command-line-arguments-microsoft-defender-antivirus.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-antivirus/command-line-arguments-microsoft-defender-antivirus.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

nam31:
Closing via #9338 
Thank you

PR #9338, opened 2021-03-19
added links , added version ,made bold sentences

RAJU2529:
as per user feedback #9333 , so i added  new version , added new links  and i made some sentences are bold


denisebmsft (approved):
Thank you @RAJU2529.

Issue #9341, opened 2021-03-19
This document recommends a different setting for “Turn off Microsoft Defender Antivirus.” than what other document suggest (the latter actually fixed issue for customer))


anaguile:
[ From RAVE case 24527130   https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/mcafee-to-microsoft-defender-setup#enable-microsoft-defender-antivirus-on-your-windows-client-devices recommends “Turn off Microsoft Defender Antivirus.” to be set to "Disabled" but https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-antivirus/use-group-policy-microsoft-defender-antivirus says it should be “Not Configured” instead to ensure any installed third-party antivirus apps work correctly, that setting is what actually fixed the problem in customer's scenario, which was Defender not going into passive mode despite being onboarded on MDE with McAfee AV installed (they provided the feedback that this should be addressed/corrected).]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 89449bef-485d-caff-f3a5-c892b3d33056
* Version Independent ID: 669feff2-5757-3c24-c803-28d6ad2f8d12
* Content: [McAfee to Microsoft Defender for Endpoint - Setup - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/mcafee-to-microsoft-defender-setup#enable-microsoft-defender-antivirus-on-your-windows-client-devices)
* Content Source: [windows/security/threat-protection/microsoft-defender-atp/mcafee-to-microsoft-defender-setup.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/microsoft-defender-atp/mcafee-to-microsoft-defender-setup.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @denisebmsft
* Microsoft Alias: **deniseb**

RAJU2529:
@anaguile . Turn off Microsoft Defender Antivirus  set to Disabled means still Microsoft Defender runs background.

Turn off Microsoft Defender Antivirus. set to Enabled means  Microsoft Defender is turned off.

RAJU2529:
PR #9343 is  complicated , i taken care while editing this article, so i request reviewers to review changed parts must be reviewed.

anaguile:
[...]

PR #9343, opened 2021-03-20
made bold sentences, removed disabled

RAJU2529:
as per user report issue #9341 so i removed disabled and added not configured. also, i made many sentences are bold.
i need  suggested commit  changes from contributors.


denisebmsft:
Thank you @RAJU2529. We don't actually bold product names so I did undo those changes. Appreciate your help addressing issues and for your pull requests.

Issue #9361, opened 2021-03-24
Extra "\" characters listed for paths to files


tomdan-MSFT:
Near the top of the article we list a path for files that include extra "\" :

The XSDs for all EAP methods are shipped in the box and can be found at the following locations:

C:\\Windows\\schemas\\EAPHost
C:\\Windows\\schemas\\EAPMethods

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c8b1b81d-3c35-0315-d149-f10af3a409b2
* Version Independent ID: c34e7acb-ece7-d76f-fb41-470e19119ee0
* Content: [VPNv2 CSP - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/mdm/vpnv2-csp)
* Content Source: [windows/client-management/mdm/vpnv2-csp.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/vpnv2-csp.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @ManikaDhiman
* Microsoft Alias: **dansimp**

RAJU2529:
@tomdan-MSFT . Thanks for pointing out.

PR #9363, opened 2021-03-24
removed backward-slash

RAJU2529:
as per user feedback #9361 , so i removed extra backward slash

### DIFF
--- a/windows/client-management/mdm/vpnv2-csp.md
+++ b/windows/client-management/mdm/vpnv2-csp.md
@@ -30,8 +30,8 @@ Here are the requirements for this CSP:
 
 The XSDs for all EAP methods are shipped in the box and can be found at the following locations:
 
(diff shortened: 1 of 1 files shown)


illfated (approved):
Good correction and update. The double backslashes stem from PR #8539 and commit https://github.com/MicrosoftDocs/windows-itpro-docs/commit/1ba86d042061cb05a5cd050abe60edfc7310c111 where MD backtick encapsulation was added, while also forgetting that the encapsulation blocks the first backslash from being read as an escape character, initially added to show one actual backslash in each instance.

JohanFreelancer9 (approved):

illfated:
@RAJU2529 : For your information: the name is "backslash" (https://en.wikipedia.org/wiki/Backslash).

Thank you for your continued work in this repository. 😃

PR #9379, opened 2021-03-29
Clarify that Stop errors are _not_ app errors

randomascii:
There is still confusion in BSOD analysis about whether a user-mode
process can cause a stop error and it is important to clarify that they
cannot.

### DIFF
--- a/windows/client-management/troubleshoot-stop-errors.md
+++ b/windows/client-management/troubleshoot-stop-errors.md
@@ -35,7 +35,8 @@ Our analysis of the root causes of crashes indicates the following:
 - 5 percent are caused by Microsoft code
 - 15 percent have unknown causes (because the memory is too corrupted to analyze)
 
-Note that the root cause of Stop errors is never a user-mode process. While a user-mode process (such as Notepad or Slack) may trigger a Stop error, it is merely exposing the underlying bug which is always in a driver, hardware, or the OS.
+>[!NOTE]
+>Note that the root cause of Stop errors is never a user-mode process. While a user-mode process (such as Notepad or Slack) may trigger a Stop error, it is merely exposing the underlying bug which is always in a driver, hardware, or the OS.
 
 ## General troubleshooting steps
 
--- a/windows/client-management/troubleshoot-stop-errors.md
+++ b/windows/client-management/troubleshoot-stop-errors.md
@@ -35,8 +35,8 @@ Our analysis of the root causes of crashes indicates the following:
 - 5 percent are caused by Microsoft code
 - 15 percent have unknown causes (because the memory is too corrupted to analyze)
 
->[!NOTE]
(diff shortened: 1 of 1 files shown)


randomascii:
This PR was suggested in a twitter discussion (https://twitter.com/ASwisstone/status/1376647562573271053) of the challenges of convincing customers that a BSOD that is triggered by our app is actually the fault of a driver/hardware/OS problem

JohanFreelancer9 (approved):

randomascii:
> Just a suggestion, based on my previous experience with MS Docs and the general codestyle therein. Would it be practical to set this additional text in a colored Note blob?
> 
> For comparison, the document contains 2 Note blobs and a Warning blob already, just in case you are wondering what it may look like.
> 
> Formatting suggestion:
> 
> ```
> > [!NOTE]
> > The root cause of Stop errors is never a user-mode process. While a user-mode process (such as Notepad or Slack) may trigger a Stop error, it is merely exposing the underlying bug which is always in a driver, hardware, or the OS.
> ```

Good idea. Done.

windows/client-management/troubleshoot-stop-errors.md
```diff
@@ -35,6 +35,9 @@ Our analysis of the root causes of crashes indicates the following:
 - 5 percent are caused by Microsoft code
 - 15 percent have unknown causes (because the memory is too corrupted to analyze)
 
+>[!NOTE]
+>Note that the root cause of Stop errors is never a user-mode process. While a user-mode process (such as Notepad or Slack) may trigger a Stop error, it is merely exposing the underlying bug which is always in a driver, hardware, or the OS.
```
illfated, line 39: While the original text starts with "Note", it seems a bit redundant to do so when there is already a title or label image above it.

May I suggest reading this note as if the word "Note" is already stated, starting the sentence like this?

```suggestion
> [!NOTE]
> The root cause of Stop errors is never a user-mode process. While a user-mode process (such as Notepad or Slack) may trigger a Stop error, it is merely exposing the underlying bug which is always in a driver, hardware, or the OS.
```
I also added MarkDown indent marker compatibility spacing (1 space).

e0i:
@randomascii Although it is common OS knowledge that the mentioned errors cannot surface from the user-mode processes, the note might be useful for the confused readers. It might take a few days to have this PR copy edited and approved. Thank you for the PR and hope you contribute to the Docs again.

randomascii:
> @randomascii Although it is common OS knowledge that the mentioned errors cannot surface from the user-mode processes, the note might be useful for the confused readers. It might take a few days to have this PR copy edited and approved. Thank you for the PR and hope you contribute to the Docs again.

Understood. Thanks for the help with the process.

illfated (approved):
Thank you for the Note adjustments. Looks very good to me.

illfated:
#sign-off

JohanFreelancer9 (approved):

PR #9389, opened 2021-04-06
Update hello-hybrid-cert-whfb-settings-dir-sync.md

DeployWindowsCom:
According to the text below, it should apply to Key trust and certificate trust

"In hybrid deployments, users register the public portion of their Windows Hello for Business credential with Azure.  Azure AD Connect synchronizes the Windows Hello for Business public key to Active Directory.

The key-trust model needs Windows Server 2016 domain controllers, which configure the key registration permissions automatically; however, the certificate-trust model does not and requires you to add the permissions manually."


illfated (approved):

JohanFreelancer9 (approved):

mapalko:
This documentation is a part of the certificate trust deployment guide section which is why it is only labelled applicable to cert trust. There is separate similar documentation in the key trust deployment guide. I do not think the applies to label should be added for key trust since this deployment guide is not for that topic.

PR #9457, opened 2021-04-26
Update policy-csp-admx-windowsexplorer.md

GenerAhl:
For the section "If you enable this setting, users can browse the directory structure of the selected drives in My Computer or File Explorer, but they cannot open folders and access the contents. Also, they cannot use the Run dialog box or the Map Network Drive dialog box to view the directories on these drives." what is the difference between "users can browse the directory structure" and "they cannot open folders"? To me, browse a directory structure includes navigating through the folders.

### DIFF
--- a/windows/client-management/mdm/policy-csp-admx-windowsexplorer.md
+++ b/windows/client-management/mdm/policy-csp-admx-windowsexplorer.md
@@ -8,7 +8,7 @@ ms.prod: w10
 ms.technology: windows
 author: manikadhiman
 ms.date: 10/29/2020
-ms.reviewer: 
+ms.reviewer:  
 manager: dansimp
 ---
 


illfated:
> "If you enable this setting, users can browse the directory structure of the selected drives in My Computer or File Explorer, but they cannot open folders and access the contents. Also, they cannot use the Run dialog box or the Map Network Drive dialog box to view the directories on these drives."

I haven't myself actually used a scenario where this setting has been applied to the operating system, but while reading this, I imagine being able to list the folders in the left hand pane of File Explorer/My Computer, maybe even see a listing of subfolders in the right hand view (not quite sure about that detail). I also imagine an optional `Tree` command view in a CMD window, but that is only a guess on my part.

It also looks a bit odd to me to start a Pull Request without having any present suggestion to review, instead of using a regular issue ticket. If I were in this situation, wanting to discuss this topic, I would have gone to the Feedback section at the bottom of the page in question, and opened a new ticket from the button link named **[:octocat: This page]**:

- https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-admx-windowsexplorer#feedback

image

illfated:
Also, to try to answer your question

>  what is the difference between "users can browse the directory structure" and "they cannot open folders"?

**Open folders** usually means that you can view the entire folder content (files, links/shortcuts, subfolders etc.) whereas **Browse the directory structure** only refers to listing subfolders and parent folders (depending on access and root folder location). There is an important logical difference between the two, which is also reflected other places like User Privileges,  Access Control, User Rights Assignment, and so on. These are key parts of Windows data access administration.

PR #9553, opened 2021-05-20
Update basic-level-windows-diagnostic-events-and-fields-1903.md

siosulli:


illfated (approved):
Fair point, addition of a sentence period dot (coherent with the lines above).

Looks good to me.

PR #9558, opened 2021-05-20
3133 url fix

I-Cat:
fix http:// to https://
Files changed 546
Assignees: MicrosoftDocs/officedocs-admin


illfated:
This PR seems to be over the top ambitious, as well as being unneeded in terms of neglecting the fact that several of the service URL addresses in the documents affected by this PR **do not support** HTTPS.

Some of the affected URLs are meant to be only for internal use, and are therefore limited to HTTP by design. Others are standardized on HTTP and do not provide their services over the HTTPS (TCP 443) protocol. There are also some non-Microsoft additions in this PR, so if I would have anything to say in this matter, I would recommend against this change.

JohanFreelancer9 (commented):
@e0i - I'll wait for clarification/review from the doc/code owner before I do a copy review. Please also see @illfated's comment.

illfated:
Also: the PR author does not appear to be a Microsoft Docs member or employee, so from my point of view, it will not be acceptable to merge material/documents where this PR author has inserted their own GitHub user name as part of the documentation.

traya1:
> This PR seems to be over the top ambitious, as well as being unneeded in terms of neglecting the fact that several of the service URL addresses in the documents affected by this PR **do not support** HTTPS.
> 
> Some of the affected URLs are meant to be only for internal use, and are therefore limited to HTTP by design. Others are standardized on HTTP and do not provide their services over the HTTPS (TCP 443) protocol. There are also some non-Microsoft additions in this PR, so if I would have anything to say in this matter, I would recommend against this change.

Closing PR for the quoted reason(s). I appreciate the effort but we have workstreams in place to that take of link issues in our topics.

I-Cat:
> This PR seems to be over the top ambitious, as well as being unneeded in terms of neglecting the fact that several of the service URL addresses in the documents affected by this PR **do not support** HTTPS.
> 
> Some of the affected URLs are meant to be only for internal use, and are therefore limited to HTTP by design. Others are standardized on HTTP and do not provide their services over the HTTPS (TCP 443) protocol. There are also some non-Microsoft additions in this PR, so if I would have anything to say in this matter, I would recommend against this change.

Yep my bad

PR #9554, opened 2021-05-20
added  Allow Update Compliance Processing

RAJU2529:
as per user feedback issue #9540,  so I added  **Allow Update Compliance Processing**   policy-related settings in this article, after looking at GPO in windows 10  pre release build 21h1 19043.985.

PR #9559, opened 2021-05-21
made boot to System/BootStartDriverInitialization

RAJU2529:
as per user feedback from @illfated under issue #9554 , so i made sentence **System/BootStartDriverInitialization**  to bold.


illfated (approved):
Looks like the correct solution to me.

Just remember that the other PR will cause a file conflict when it gets merged, these lines will become 925-930 instead of 855-860.

JohanFreelancer9 (approved):

PR #9565, opened 2021-05-21
Update sign-code-integrity-policy-with-device-guard-signing.md

linddief:
Update DGSSv1 retirement date.


illfated (approved):
Looks good to me, at least as far as the changes go. Maybe there could be something else in the unchanged parts of the text, that JohanFreelancer9 would like to offer improvements to, but we'll see if he offers any further changes.

PR #9627, opened 2021-05-31
removed see also  added related topics

RAJU2529:
this is my own PR, I removed see also  and added Related topics

### DIFF
--- a/windows/client-management/mdm/vpnv2-csp.md
+++ b/windows/client-management/mdm/vpnv2-csp.md
@@ -1592,7 +1592,7 @@ Servers
       </Add>
 ```
 
-## See also
+## Related topics
 
 [Configuration service provider reference](configuration-service-provider-reference.md)
 


windows/client-management/mdm/vpnv2-csp.md
```diff
@@ -1592,7 +1592,7 @@ Servers
       </Add>
 ```
 
-## See also
+## Related topics
```
illfated, line 1595: @RAJU2529 : Could you take me through your reasoning for this change?

From my point of view, it does not appear to be necessary here, because the following link points to a reference document page, containing the specific reference manual to this current page (https://docs.microsoft.com/en-us/windows/client-management/mdm/vpnv2-csp#see-also)

I know, it may be the case that "Related topics" is used a lot more ([733 code results in MicrosoftDocs/windows-itpro-docs](https://github.com/MicrosoftDocs/windows-itpro-docs/search?q=%22Related+topics%22)) compared to "See also", which is used only 85 times ([85 code results in MicrosoftDocs/windows-itpro-docs](https://github.com/MicrosoftDocs/windows-itpro-docs/search?q=%22See+also%22)), but in this specific case the current wording seems to be the most appropriate phrasing as it already is.

illfated: Admittedly, I am not the real expert here, so I defer to the MS Docs team as well as their tech writers and copy review experts.

JohanFreelancer9 (approved):

e0i:
@RAJU2529 

Unfortunately, the PR is missing substance and does not address a specific problem in the documentation. Please do ensure to come up with the rationale the next time.

Thank you for your ongoing efforts to make the documentation accurate, and please stay safe.

Issue #9636, opened 2021-06-01
Configure the auto-enrollment for a group of devices - Important banner


MicahFalde:
[Enter feedback here]
It is not clear in this "Important" section whether you are looking for the GPO in the .admx template on the DC, or when checking whether or not it applied with "gpresult /r" on the machine you are working with.

"If you do not see the policy".  If I don't see the policy where?  If I don't see the policy in the .admx template when I am going to set it up on the DC?  In my situation, I do see that.  However, I do not see the policy applying on the endpoint when checking for it.  So I don't know whether or not this "Important" is applicable to me or not.

Here is the screenshot of the section I am talking about.  This is under the heading "Configure the auto-enrollment for a group of devices"
![image](https://user-images.githubusercontent.com/17536628/120391789-2c3e7980-c2f5-11eb-9240-0fc81c759f86.png)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ad012591-5c74-f1cd-d3a2-443768e8ceee
* Version Independent ID: 918ea4bc-0849-3745-5d9d-a21a973bcd3b
* Content: [Enroll a Windows 10 device automatically using Group Policy - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy#feedback)
* Content Source: [windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/enroll-a-windows-10-device-automatically-using-group-policy.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @ManikaDhiman
* Microsoft Alias: **dansimp**

PR #9642, opened 2021-06-02
added  may 2021 admx link

RAJU2529:
as per user feedback #9636 ,  so i added  may 2021 admx template link.


denisebmsft (approved):
Thank you @RAJU2529

PR #9681, opened 2021-06-14

pelarsen:
Device Credential
Is only supported for:
- Co-management
- WVD (Azure Virtual Desktop)
- Autopilot deploying mode - witch is not using GPO for enrollment


JohanFreelancer9 (approved):
Copy review.

PR #9692, opened 2021-06-16
Updated error AllowWufbCloudProcessing

cinglis-msft:
Should be value 8, showed value 1


Ar5ham (approved):

windows/deployment/update/deployment-service-overview.md
```diff
@@ -125,7 +125,7 @@ Deployment scheduling controls are always available, but to take advantage of th
 > Deployment protections are currently in preview and available if you're using Update Compliance. If you set these policies on a a device that isn't enrolled in Update Compliance, there is no effect.
```
JohanFreelancer9, line 125:
```suggestion
> Deployment protections are currently in preview and available if you're using Update Compliance. If you set these policies on a device that isn't enrolled in Update Compliance, there is no effect.
```

JohanFreelancer9 (changes requested):
@cinglis-msft - Could you please also commit my suggested change to remove the double "a" from line 125? Many thanks.

PR #9723, opened 2021-06-24
fixed grammar in Windows 11 requirements.

Wmlr:
"an HDR monitor".
If the letter H is pronounced you should be using "a HDR monitor"


JohanFreelancer9 (approved):

greg-lindsay (commented):
Hi, this change is incorrect.  The correct article is "an" here.  You use "an" when the next word sounds like a vowel, not when the next word is a vowel.

PR #9762, opened 2021-06-30
Update system-guard-secure-launch-and-smm-protection.md

brentkendall:
Clarified that some Intel fTPMs (technically "integrated" TPMs) are now supported.

Intel has had an integrated hardware TPM in its CPUs for a while.  It is not usually enabled on.  They call it Platform Trust Technology or PTT, and it meets the TPM 2.0 spec.  OEMs can use this PTT instead of a discrete TPM.  The Core i5-10310U is one of these Comet Lake chips with PTT.


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @brentkendall

Issue #9766, opened 2021-07-01
Minor typo


Rvvhub:
Minor typo: 

"To limit the number of changed domain credentials "

should be:

"To limit the number of cached domain credentials "

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e3ded1fc-7605-b135-d643-fdc2f40fbb0c
* Version Independent ID: 3e5a97a6-4557-19a1-9980-fbc602251573
* Content: [Network access Do not allow storage of passwords and credentials for network authentication (Windows 10) - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/network-access-do-not-allow-storage-of-passwords-and-credentials-for-network-authentication)
* Content Source: [windows/security/threat-protection/security-policy-settings/network-access-do-not-allow-storage-of-passwords-and-credentials-for-network-authentication.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/security-policy-settings/network-access-do-not-allow-storage-of-passwords-and-credentials-for-network-authentication.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

RAJU2529:
@Rvvhub . thanks for pointing out.

denisebmsft:
Than you @Rvvhub for letting us know about this, and thank you @RAJU2529 for fixing it. The update will be live a little later this morning.

PR #9767, opened 2021-07-01
typo correction

RAJU2529:
as per user report #9766, so i corrected the word from **changed** to **cached**


denisebmsft (approved):
Thank you @RAJU2529

Issue #9768, opened 2021-07-01
Spelling error


markburland:
"Because of this, you can no longer manually set the service Startup type to Automatic by using the **Sevices** snap-in."

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5118f61c-e134-82dc-bde9-12dc1e54a5e5
* Version Independent ID: cfa8bb11-4c14-fd94-2988-7b658e4a017f
* Content: [Configure the Application Identity service (Windows 10) - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/windows-defender-application-control/applocker/configure-the-application-identity-service)
* Content Source: [windows/security/threat-protection/windows-defender-application-control/applocker/configure-the-application-identity-service.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/windows-defender-application-control/applocker/configure-the-application-identity-service.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

RAJU2529:
@markburland , Thanks for pointing out.

denisebmsft:
Hello @markburland and thank you for letting us know about this. Thank you @RAJU2529 for the fix. The update will be live a little later this morning.

PR #9769, opened 2021-07-01
typo correction

RAJU2529:
as per user report issue #9768 , so i corrected  **sevices**  to  **services**
Also i conformed this article is applies to  Windows 11 too


denisebmsft (approved):
thank you @RAJU2529

Issue #9782, opened 2021-07-05
Incorrect path for registry location for SetupDiag.exe


brookspeppin:
The registry location documented for where SetupDiag.exe stores the results is incorrect under the heading SetupDiag in Windows 10, version 2004 and later. _"The resulting SetupDiag analysis can be found at %WinDir%\Logs\SetupDiag\SetupDiagResults.xml and in the registry under HKLM\SYSTEM\Setup\SetupDiag\Results."_

This is incorrect. It was changed in version v1.5.0.0 to store them here:
HKLM\SYSTEM\Setup\MoSetup\Volatile\SetupDiag

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 5ad81361-15d8-0e95-9d5f-090dbf270f10
* Version Independent ID: c0c32397-d5d6-cce1-79fd-c310bec05240
* Content: [SetupDiag - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/upgrade/setupdiag)
* Content Source: [windows/deployment/upgrade/setupdiag.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/upgrade/setupdiag.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @greg-lindsay
* Microsoft Alias: **greglin**

RAJU2529:
@denisebmsft , Sir in this article , Tools folder must be removed C;\

greg-lindsay:
The tools folder doesn't need to be removed. The path is arbitrary. It could be z:\temp or anything. This is merely a location to manually save files.  This also doesn't address the initial comment which refers to automatic file saving path.

greg-lindsay:
With respect to the original comment: there is a different path used when SetupDiag is run manually, vs. when it is run automatically by Windows Setup.  If you read the article carefully you will see that when setupdiag is run automatically, the regpath parameter is used which changes the registry location for log files.  

I am adding a note on the article so that this is a little more clear.

PR #9783, opened 2021-07-06
removed tools folder

RAJU2529:
as per user feedback #9782 , after reading this article, I found **Tools** folder must be removed from   C:\ , so I removed  **C:\Tools**


JohanFreelancer9 (approved):

greg-lindsay:
No this is not correct. The path is arbitrary.

PR #9840, opened 2021-07-23
Update defender-csp.md

marysiakam9889:
Corrected signature to definition in tree format


denisebmsft (approved):
Thank you @marysiakam9889

Issue #9824, opened 2021-07-20
Can i add powershell commands


RAJU2529:
Can i add PowerShell commands for the following topic

To enable VBS with Secure Boot and DMA (value 3), in the preceding command, change /d 1 to /d 3.

To enable VBS with UEFI lock (value 1), in the preceding command, change /d 0 to /d 1.

To enable VBS with Secure Boot and DMA (value 3), in the preceding command, change /d 1 to /d 3.

To enable virtualization-based protection of Code Integrity policies with UEFI lock (value 1), in the preceding command, change /d 0 to /d 1.

To enable VBS with Secure Boot and DMA (value 3), in the preceding command, change /d 1 to /d 3.

#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1d4b8d75-5d09-e0d6-1a90-45175fd1043d
* Version Independent ID: dfb10d10-a514-6ea5-1307-df865db530db
* Content: [Enable virtualization-based protection of code integrity - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/device-guard/enable-virtualization-based-protection-of-code-integrity)
* Content Source: [windows/security/threat-protection/device-guard/enable-virtualization-based-protection-of-code-integrity.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/device-guard/enable-virtualization-based-protection-of-code-integrity.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

RAJU2529:
@e0i I successfully added  more commands

e0i:
Thank you, @RAJU2529

PR #9859, opened 2021-07-30
Added powershell commands

RAJU2529:
as per user feedback #9824, so i added correct commands after verification from the windows 11 dev channel.

Also, I found that TWO screenshots are really old, so I want to add new screenshots.


denisebmsft (approved):
Thank you @RAJU2529

PR #9896, opened 2021-08-13
Removing ALLOW_ALL for FileRules

carlosmayol:
These rules are enabling any executable regardless of the signers policies.


JohanFreelancer9 (approved):

denisebmsft (approved):
Thank you @carlosmayol

Issue #9934, opened 2021-09-01
Using the phrase "PKI Infrastructure" is redundant. Please correct.


nadcraker:
Any time in this page that you say "PKI Infrastructure" please just say "PKI." The "I" in "PKI" already means infrastructure. The way this article uses the term is incorrect.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 94cebf1d-b16d-4ddc-b532-be829b858938
* Version Independent ID: 3707f457-98d5-e5b3-c749-99afe0d6d495
* Content: [4768(S, F) A Kerberos authentication ticket (TGT) was requested. (Windows 10) - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4768)
* Content Source: [windows/security/threat-protection/auditing/event-4768.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/auditing/event-4768.md)
* Product: **m365-security**
* Technology: **mde**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

RAJU2529:
@nadcraker Thanks for pointing out. I will correct today afternoon.

PR #9935, opened 2021-09-03
removed word

RAJU2529:
as per user feedback #9934 , so i removed the word **Infrastructure**


denisebmsft (approved):
Thank you @RAJU2529

PR #9968, opened 2021-09-21
Make Domain Admins well-known SID consistent with others

cnotin:
It was missing the "-21-" part which all other similar well-known have.
For example, see just below: "Domain Computers" -> "S-1-5-21-<domain>-515


denisebmsft (approved):
thank you @cnotin

PR #9969, opened 2021-09-21
Enterprise Read Only Domain Controllers (-498) are defined at forest root level

cnotin:
Syntax is similar for example to "Enterprise Admins" -> "S-1-5-21-\<root domain\>-519"

Proof [number 1](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/5b0b345f-ff18-4058-9f7c-76431e514f50):
> This group is created in the root domain by the root domain 

Proof [number 2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/81d92bba-d22b-4a8c-908a-554ab29148ab):
> ENTERPRISE_READONLY_DOMAIN_CONTROLLERS
> S-1-5-21-\<root domain\>-498


denisebmsft (approved):
thank you @cnotin

PR #10013, opened 2021-10-07
Update on Network Security Pk2U Mitigation/Fix

fmata-ms:
Added the mitigation information and link to the proper CVE for easy identification so that the users can easily understand how to address it if no longer an issue.


aczechowski (approved):
Thanks @fmata-ms for adding this info, and thank you for your first contribution to the Windows repository! 🎉

PR #10081, opened 2021-11-07
Updated DC certificate subject name requirement

ajf8729:
The subject name of the DC certificate should be "fully distinguished name", not "none", according to https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-hybrid-key-trust-prereqs#public-key-infrastructure

> "The certificate Subject section should contain the directory path of the server object (the distinguished name)."


JohanFreelancer9 (approved):

mapalko:
This instruction should not be changed.

ccatlett1984:
> This instruction should not be changed.

The docs are conflicting, one of them must be wrong.

aczechowski:
I'm checking on this

paolomatarazzo:
Hello @ajf8729, the docs are not conflicting, since they are using the words "should" or "could". The fact that the DC certs are required to have a SAN "The certificate Subject Alternative Name section must contain the Domain Name System (DNS) name", it implies that the subject name is not used, therefore it's irrelevant. Thank you for the feedback!

PR #10102, opened 2021-11-11
reemoved long title sentences

RAJU2529:
this is my own PR,  we could not able to identify adjustment is correct or not, before  creating PR  and before publishing. so we have to take a test drive.


JohanFreelancer9 (approved):

aczechowski:
@RAJU2529 what's the intended result here? It looks like you removed most of the bullets from the top of the article, which I think is a great idea. The docs.ms platform creates links for every H2 header as a part of the "In this article" menu to the right, so don't need to manually maintain a list. But there's still one bullet on line 25, is that intended?

aczechowski:
OK, I'm making the assumption that the extra line should be removed. Merging this change now

PR #10167, opened 2021-11-30
added applies to

RAJU2529:
after reading this  article, I added windows 11 and windows 10  under  applies to section


denisebmsft (approved):
thank you @RAJU2529

Issue #10227, opened 2021-12-21
wrong gp name! 


attaya:
below description is wrong:

GP Friendly name: Allow time zone redirection
GP name: TS_GATEWAY_POLICY_ENABLE
GP path: Windows Components\Remote Desktop Services\Remote Desktop Session Host\Device and Resource Redirection
GP ADMX file name: TerminalServer.admx

the right name for "Allow time zone redirection" is TS_TIME_ZONE! 

it causes confusing to our customer, i am a CSS engineer. 

[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 2d881506-cf8c-4e25-16cd-df477ed6fb0c
* Version Independent ID: 98158dde-730b-ec77-d2f9-c5e045d69dad
* Content: [Policy CSP - ADMX_TerminalServer - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/mdm/policy-csp-admx-terminalserver)
* Content Source: [windows/client-management/mdm/policy-csp-admx-terminalserver.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/policy-csp-admx-terminalserver.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

MaratMussabekov:
Hello @Dansimp ,

There are two different ADMX templates exist - terminalserver-server.admx and terminalserver.admx
There is some confusion related to the article **Policy CSP - ADMX_TerminalServer**. Its name sounds like it replicates terminalserver.admx template, however the description of the policies are taken from terminalserver-server.admx policies (from policies "Allow time zone redirection" and "Do not allow Clipboard redirection").
Meanwhile, the policy names and the ADMX info section from the policies are taken from terminalserver.admx (from policies "Enable connection through RD Gateway" and "Set RD Gateway authentication method")

Thank you

vinaypamnani-msft:
Fixed in PR #10229

PR #10229, opened 2021-12-21
corrected word

RAJU2529:
as per user report #10227 , so i corrected, after verifying with windows 11 build no 22000.376 admx templates.

Screenshot 2021-12-21 174920


denisebmsft (approved):
Thank you @RAJU2529

Issue #10272, opened 2022-01-11
Command spelling incorect


abhishekjha1990:
Hi --Command mentioned for 0x800f0825 code spell wrong 'DISM /ONLINE /CLEANUP-IMAGE /CHECKHEALT'
please correct to DISM /ONLINE /CLEANUP-IMAGE /CHECKHEALTH

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1cb9f2a2-76ad-8301-7d80-e33a7bb80a1b
* Version Independent ID: dd070c32-53d5-53d6-7ba2-a4361b5263bb
* Content: [Windows Update common errors and mitigation - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/update/windows-update-errors)
* Content Source: [windows/deployment/update/windows-update-errors.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/update/windows-update-errors.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @jaimeo
* Microsoft Alias: **greglin**

jaimeo:
Using the "Edit" button you can make this correction yourself. We'll happily merge it. Thanks!

RAJU2529:
@abhishekjha1990 . I will fix this issue evening after finishing my college work.

RAJU2529:
@jaimeo , Please verify PR #10275 and merge if correct.

PR #10275, opened 2022-01-12
typo correction

RAJU2529:
as per user report #10272, so i changed  **HEALT** to  **HEALTH
adjusted  **Microsoft update servers**  address


jaimeo:
FYI: I don't work in this area anymore. A new person should be changing the authorship on these files pretty soon.

PR #10293, opened 2022-01-18
Removing groups option from Windows 10 version 2004

secherka:
After working the case and talking to PG, Ravi Vennapusa, confirmed we cant add groups to Remote Desktop Users. Only users can be added as mentioned in the linked in this section article. Also submitting change request to this branch, since not able to access Review one, getting 404 error.


denisebmsft (approved):
Thank you @secherka

PR #10348, opened 2022-02-07
Fix typos

GitHubPang:


windows/client-management/troubleshoot-tcpip-port-exhaust.md
```diff
@@ -24,7 +24,7 @@ There are two types of ports:
...
  
-In a scenario where the same browser is creating a lot of connections to multiple website, for any new connection that the browser is attempting, an ephemeral port is used. After some time, you will notice that the connections will start to fail and one high possibility for this would be because the browser has used all the available ports to make connections outside and any new attempt to establish a connection will fail as there are no more ports available. When all the ports are on a machine are used, we term it as *port exhaustion*. 
+In a scenario where the same browser is creating a lot of connections to multiple website, for any new connection that the browser is attempting, an ephemeral port is used. After some time, you will notice that the connections will start to fail and one high possibility for this would be because the browser has used all the available ports to make connections outside and any new attempt to establish a connection will fail as there are no more ports available. When all the ports on a machine are used, we term it as *port exhaustion*.
```
JohanFreelancer9, line 27:
```suggestion
In a scenario where the same browser is creating a lot of connections to multiple websites, for any new connection that the browser is attempting, an ephemeral port is used. After some time, you will notice that the connections will start to fail and one high possibility for this would be because the browser has used all the available ports to make connections outside and any new attempt to establish a connection will fail as there are no more ports available. When all the ports on a machine are used, we term it as *port exhaustion*. 
```

JohanFreelancer9 (changes requested):
Copy review - @GitHubPang, kindly commit the additional suggested change. Many thanks.

JohanFreelancer9 (approved):

denisebmsft (commented):
Thank you @GitHubPang @JohanFreelancer9

PR #10374, opened 2022-02-22
Update HideExclusionsFromLocalAdmins scope

JesseEsquivel:
Added registry to the scope or where the exclusions are hidden from (as this is the case).


denisebmsft (approved):
Thank you @JesseEsquivel

Issue #10328, opened 2022-01-31
This page is dangerous


Vroo:
[Enter feedback here]
To most readers PINs mean a short code like they use to unlock their phone or use with a debit / ATM card. Furthermore, the point you are making is that passwords stored locally are safer than those stored on a server. However many sites use what they call PINs as server passwords.

Categorically stating that PINs are better is dangerous.

All of the benefits you claim for PINs apply to local passwords and PINs and all the disadvantages of passwords apply to server passwords and PINs.

If what you are saying is that Windows has a defect that it doesn't secure local passwords with the TPM, then that should be made explicit here and you should link to the issue for that defect so people can find out when it is fixed.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: c34dbafa-df3c-a5a5-32d5-2ce152893534
* Version Independent ID: cebdd9a6-45d7-6306-373c-7d31121f9e16
* Content: [Why a PIN is better than a password (Windows) - Windows security](https://docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/hello-why-pin-is-better-than-password)
* Content Source: [windows/security/identity-protection/hello-for-business/hello-why-pin-is-better-than-password.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/identity-protection/hello-for-business/hello-why-pin-is-better-than-password.md)
* Product: **m365-security**
* GitHub Login: @mapalko
* Microsoft Alias: **mapalko**

PR #10401, opened 2022-03-08
Update hello-why-pin-is-better-than-password.md

michaelAngeloEgypt:
implementing suggestions given in this task
https://github.com/MicrosoftDocs/windows-itpro-docs/issues/10328

### DIFF
--- a/windows/security/identity-protection/hello-for-business/hello-why-pin-is-better-than-password.md
+++ b/windows/security/identity-protection/hello-for-business/hello-why-pin-is-better-than-password.md
@@ -27,7 +27,7 @@ ms.date: 10/23/2017
 - Windows 11
 
 Windows Hello in Windows 10 enables users to sign in to their device using a PIN. How is a PIN different from (and better than) a local password?
-On the surface, a PIN looks much like a password. A PIN can be a set of numbers, but enterprise policy might allow complex PINs that include special characters and letters, both upper-case and lower-case. Something like **t758A!** could be an account password or a complex Hello PIN. It isn't the structure of a PIN (length, complexity) that makes it better than an online password, it's how it works. First we need to distinguish between two types of passwords: 'local' passwords are validated against the machine's password store, whereas 'online' passwords are validated against a server. This article mostly covers the benefits a PIN has over an online password, and also why it can be considered even better than a local password. 
+On the surface, a PIN looks much like a password. A PIN can be a set of numbers, but enterprise policy might allow complex PINs that include special characters and letters, both upper-case and lower-case. Something like **t758A!** could be an account password or a complex Hello PIN. It isn't the structure of a PIN (length, complexity) that makes it better than an online password, it's how it works. First we need to distinguish between two types of passwords: `local` passwords are validated against the machine's password store, whereas `online` passwords are validated against a server. This article mostly covers the benefits a PIN has over an online password, and also why it can be considered even better than a local password. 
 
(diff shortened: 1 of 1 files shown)


windows/security/identity-protection/hello-for-business/hello-why-pin-is-better-than-password.md
```diff
@@ -19,42 +19,44 @@ ms.localizationpriority: medium
...
+Windows Hello in Windows 10 enables users to sign in to their device using a PIN. How is a PIN different from (and better than) a local password?
+On the surface, a PIN looks much like a password. A PIN can be a set of numbers, but enterprise policy might allow complex PINs that include special characters and letters, both upper-case and lower-case. Something like **t758A!** could be an account password or a complex Hello PIN. It isn't the structure of a PIN (length, complexity) that makes it better than an online password, it's how it works. First we need to distinguish between two types of passwords: 'local' passwords are validated against the machine's password store, whereas 'online' passwords are validated against a server. This article mostly covers the benefits a PIN has over an online password, and also why it can be considered even better than a local password.
```
JohanFreelancer9, line 30:
```suggestion
On the surface, a PIN looks much like a password. A PIN can be a set of numbers, but enterprise policy might allow complex PINs that include special characters and letters, both upper-case and lower-case. Something like **t758A!** could be an account password or a complex Hello PIN. It isn't the structure of a PIN (length, complexity) that makes it better than an online password, it's how it works. First we need to distinguish between two types of passwords: `local` passwords are validated against the machine's password store, whereas `online` passwords are validated against a server. This article mostly covers the benefits a PIN has over an online password, and also why it can be considered even better than a local password. 
```

windows/security/identity-protection/hello-for-business/hello-why-pin-is-better-than-password.md
```diff
@@ -19,42 +19,44 @@ ms.localizationpriority: medium
...
 
 >[!NOTE]
 >For details on how Hello uses asymetric key pairs for authentication, see [Windows Hello for Business](hello-overview.md#benefits-of-windows-hello).
  
 ## PIN is backed by hardware
 
-The Hello PIN is backed by a Trusted Platform Module (TPM) chip, which is a secure crypto-processor that is designed to carry out cryptographic operations. The chip includes multiple physical security mechanisms to make it tamper resistant, and malicious software is unable to tamper with the security functions of the TPM. Many modern devices have TPM.
+The Hello PIN is backed by a Trusted Platform Module (TPM) chip, which is a secure crypto-processor that is designed to carry out cryptographic operations. The chip includes multiple physical security mechanisms to make it tamper resistant, and malicious software is unable to tamper with the security functions of the TPM. Many modern devices have TPM. Windows 10, on the other hand, has a defect of not linking local passwords to TPM. This is the reason why PINs are considered more secure than local passwords.
```
JohanFreelancer9, line 53:
```suggestion
The Hello PIN is backed by a Trusted Platform Module (TPM) chip, which is a secure crypto-processor that is designed to carry out cryptographic operations. The chip includes multiple physical security mechanisms to make it tamper resistant, and malicious software is unable to tamper with the security functions of the TPM. Many modern devices have TPM. Windows 10, on the other hand, does not link local passwords to TPM. This is the reason why PINs are considered more secure than local passwords.
```

JohanFreelancer9 (changes requested):
Copy review - @michaelAngeloEgypt, kindly commit the minor suggested changes. Many thanks.

PR #10421, opened 2022-03-14
Remove private preview link from MCC Docs

vakkinep:
Private Preview link should not be public facing. Anyone with this link will be allowed to create the MCC Azure Resource


vakkinep:
@JohanFreelancer9 @greg-lindsay Can you please take a quick look at this? We need to make this change ASAP

greg-lindsay:
This PR needs to be reverted. The changes in the private repo should be kept.

Issue #10322, opened 2022-01-26
Missing enum values for Detections/ThreatId/ExecutionStatus


eppsilon:
This document lists a `Detections/ThreatId/ExecutionStatus` node but does not say what the possible enum values are. Is it the same enum as [MPEXECUTION_STATUS](https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/lwef/mpexecution-status.md)?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4f65802b-f3e2-1c15-d972-b23e53ed7e72
* Version Independent ID: cc29ae16-e01e-aad1-3768-d6b8a0b60b1a
* Content: [Defender CSP - Windows Client Management](https://docs.microsoft.com/en-us/windows/client-management/mdm/defender-csp)
* Content Source: [windows/client-management/mdm/defender-csp.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/defender-csp.md)
* Product: **w10**
* Technology: **windows**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

PR #10433, opened 2022-03-17
Update defender-csp.md as per feedback IssueId 2235

michaelAngeloEgypt:
I searched for confirmations of the feedback online and found this link stating that a status exists with the name 'Blocked', so the commenter seems right in the link he sent as source of the enum
https://support.huntress.io/hc/en-us/articles/4414493304595-Managed-AV-What-is-an-Incident-Report-

### DIFF
--- a/windows/client-management/mdm/defender-csp.md
+++ b/windows/client-management/mdm/defender-csp.md
@@ -229,6 +229,14 @@ Information about the execution status of the threat.
 
 The data type is integer.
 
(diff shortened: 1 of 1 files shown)


JohanFreelancer9 (approved):

denisebmsft:
Requesting review from @marysiakam9889

e0i:
Addresses #10322

PR #10488, opened 2022-04-21
Add a cloud environment to the Requirements:

junhasems:


JohanFreelancer9 (approved):

aczechowski (approved):
Thanks @junhasems for the addition, and thank you for your first contribution to the Windows docs content! 🎉 

In the future, as a Microsoft employee, I suggest using the private repository. In this case, it's windows-docs-pr instead of windows-itpro-docs. You can either manually change the repo name in the URL, or from the article, add "review." before docs in the URL. For example, `https://review.docs.microsoft.com/en-us/windows/security/identity-protection/hello-for-business/feature-multifactor-unlock?branch=main`
(It will automatically append the `?branch=main` variable.)
Then when you select Edit, it will automatically go to the private repository.
Let me know if you have any questions, I'm always happy to help contributors! 😄

PR #10507, opened 2022-04-29
enroll-a-windows-10-device-using-GPO

ChadSimmons:
Clarify why enrollment via Device Credential is only supported by Co-Management or AVD.
Reference to Security & Management Elite Partner forum post.


denisebmsft (approved):
Thank you @ChadSimmons

PR #10529, opened 2022-05-09
Update windowsautopilot-csp.md

cbrito01:
It appears that the string "Because the CSP description should be more general/high level" was not removed from the original template.


denisebmsft:
@nimishasatapathy can you take a look at this pull request? Seems like we might be missing context for the proposed change which actually seems like a grammar issue. Thank you!

windows/client-management/mdm/windowsautopilot-csp.md
```diff
@@ -18,7 +18,7 @@ ms.date: 02/07/2022
 > Some information relates to prereleased product which may be substantially modified before it's commercially released. Microsoft makes no warranties, express or implied, with respect to the information provided here.
 
 
-The WindowsAutopilot CSP collects hardware information about a device and formats it into a BLOB. This BLOB is used as input for calling Windows Autopilot Service to mark a device as remediation required if the device underwent a hardware change that affects its ability to use Windows Autopilot.” with “The WindowsAutopilot CSP exposes Windows Autopilot related device information.” Because the CSP description should be more general/high level.
+The WindowsAutopilot CSP collects hardware information about a device and formats it into a BLOB. This BLOB is used as input for calling Windows Autopilot Service to mark a device as remediation required if the device underwent a hardware change that affects its ability to use Windows Autopilot.” with “The WindowsAutopilot CSP exposes Windows Autopilot related device information.”
```
JohanFreelancer9, line 21:
```suggestion
The WindowsAutopilot CSP collects hardware information about a device and formats it into a BLOB. This BLOB is used as input for calling Windows Autopilot Service to mark a device as remediation required if the device underwent a hardware change that affects its ability to use Windows Autopilot.
```

JohanFreelancer9, line 21:
```suggestion
The WindowsAutopilot CSP collects hardware information about a device and formats it into a BLOB. This BLOB is used as input for calling Windows Autopilot Service to mark a device as remediation required if the device underwent a hardware change that affects its ability to use Windows Autopilot. 
```

JohanFreelancer9 (changes requested):
Copy review - @cbrito01 not sure, but I think the preceding string should also be removed because as it stands, it does not make sense. If you agree, kindly commit the suggested change. Many thanks.

cbrito01:
Agreed that the preceding text does not make sense. Thanks for reviewing that.

cbrito01:
What else is missing for these changes to be committed?

aczechowski:
Thanks @cbrito01 for catching that, and thank you for your first contribution to Windows docs! 🎉 
It looks like someone else made a similar change as well, but I fixed the merge conflict. I'll merge this now so you'll get the contribution credit 😉

PR #10712, opened 2022-07-17
PINComplexity - Minimum PIN length

mikurii:
I visited "https://endpoint.microsoft.com/#blade/Microsoft_Intune_DeviceSettings/DevicesWindowsMenu/windowsEnrollment" to configure "Windows Hello for Business" setting and found out that the default value of "Minimum PIN length" is 6 letters now. I guess the default value has been changed from 4 to 6.


JohanFreelancer9 (approved):
Copy review.

aczechowski (approved):
Thanks @mikurii for the edit, and thank you for your first contribution to windows-itpro-docs! 🎉  I tweaked the sentence to use the same value.

PR #10720, opened 2022-07-20
Update vpn-security-features.md

sazankha:


aczechowski (approved):
Thanks @sazankha for the contribution! Vinay will follow up with some additional notes. I made a few editorial tweaks here, but otherwise looks good.

PR #10722, opened 2022-07-21
Update mcc-enterprise.md

TheITCloudGuy:
Clarification for Intune Configuration setting.  Current doc doesn't indicate to the user that they're required to point the profile to the EFLOW machine and NOT the Windows Server.


aczechowski (approved):
Thanks @TheITCloudGuy for the edit, and thank you for your first contribution to windows-itpro-docs! 🎉

PR #10731, opened 2022-07-27
Update info on unenroll

hathin:
Made changes to the Microsoft/customer responsibility section. Updated a few links as well (deregister a device) @tiaraquan FYI


tiaraquan:
As per discussion with Harman, closing this PR. Made the same changes in the PR #6849.

PR #10746, opened 2022-08-01
GPO update

NagaCSC:
- update about [Interactive logon Do not require CTRL+ALT+DEL (Windows 10) - Windows security | Microsoft Docs ]


NagaCSC (commented):
updated

paolomatarazzo:
Hello @NagaCSC , thank you for the feedback. I'm not entirely sure I understand your point. Once WHFB is provisioned, it becomes the default credential provider. If you have a biometric-capable device, you will use biometric as primary method for WHFB, with fall back to PIN in case of issues. The password would be a last-resort option, and you would need to explicitly select the password provider to log in with a password. I would argue that it's harder to spoof than CTRL-ALT-DELETE. Besides, if you change your provider to password, at the next login it will default again to WHFB. Can you please clarify your statement to ensure I didn't miss anything?

NagaCSC:
updated again, please review

paolomatarazzo:
Closing this PR. After reviewing latest recommendations and security baselines, we should not recommend enabling CAD to begin with.

Issue #10390, opened 2022-03-04
Password Max Length Guidance is No Longer Sufficient


DenkertM:
As of March 2022, this page no longer meets the guidlines of NIST SP 800-62 Paragraph 5.1.1.2. 

This guidance states "Verifiers SHALL require subscriber-chosen memorized secrets to be at least 8 characters in length. Verifiers SHOULD permit subscriber-chosen memorized secrets at least 64 characters in length."

It is recommended that the guidance from Microsoft is updated to reflect the NIST guidance.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 9bcaf30b-7304-6874-ba5d-791cc4d435d2
* Version Independent ID: 961def9f-094a-3d1e-82a4-bb152c2e5a1e
* Content: [Minimum password length (Windows 10) - Windows security](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/minimum-password-length)
* Content Source: [windows/security/threat-protection/security-policy-settings/minimum-password-length.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/security-policy-settings/minimum-password-length.md)
* Product: **m365-security**
* Technology: **windows-sec**
* GitHub Login: @Dansimp
* Microsoft Alias: **dansimp**

aczechowski:
Thanks @DenkertM for the feedback. I'm checking with the content owner, as it looks like there's been some previous discussions around this setting. For example, see #6947 (which references several others).

paolomatarazzo:
@DenkertM thank you for your contribution. Can you please elaborate how the article is not reflecting the NIST guidelines?
I just updated the article with "Set Minimum password length to at least a value of 8". I'm not sure if you were referring to that sentence.

✅ "Verifiers SHALL require subscriber-chosen memorized secrets to be at least 8 characters in length"

Windows also allows passwords longer that 64 characters in length, so it fulfills the second NIST requirement
✅"Verifiers SHOULD permit subscriber-chosen memorized secrets at least 64 characters in length."

The article that you are pointing out, is about Minimum password length.

PR #10756, opened 2022-08-06
Update minimum-password-length.md

saif71:
Issue: #10390 

Two changes:
• In the Best Practice section:
1. Changed the capitalization from "Set Minimum..." => "Set minimum..."
2. In the first line it says "**Set minimum password length to at least a value of 14**", but on the next line it says "**In most environments, an eight-character password is recommended because it's long enough...**"; a little bit contradictory with each other.  So I changed the minimum password length to 8. 
Also, as of Aug 2022, this is suggested in the **guidelines of NIST SP 800-62 Paragraph 5.1.1.2.**

Ref: https://nvlpubs.nist.gov/nistpubs/specialpublications/nist.sp.800-63b.pdf (Page: 23, Section: 5.1.1.2)

### DIFF
--- a/windows/security/threat-protection/security-policy-settings/minimum-password-length.md
+++ b/windows/security/threat-protection/security-policy-settings/minimum-password-length.md
@@ -36,7 +36,7 @@ The **Minimum password length** policy setting determines the least number of ch
 
 ### Best practices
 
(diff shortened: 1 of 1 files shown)


JohanFreelancer9 (approved):

aczechowski:
Thanks @saif71 for the edit. I'm checking with the content owner, as it looks like there's been some previous discussions around this setting. For example, see #6947 (which references several others).

paolomatarazzo:
Thank you for your contribution, @saif71. I approved the proposed change.

PR #10863, opened 2022-09-16
Update introduction-vamt.md

baardhermansen:
## Why

Update with correct information. 

## Changes

Updated text regarding OS that VAMT can be installed on, so it's in line with information on volume-activation-management-tool.md. 
Added a hard line shift in text under KMS section, since that seems like the authors intention.

### DIFF
--- a/windows/deployment/volume-activation/introduction-vamt.md
+++ b/windows/deployment/volume-activation/introduction-vamt.md
@@ -12,7 +12,7 @@ ms.topic: article
 
 # Introduction to VAMT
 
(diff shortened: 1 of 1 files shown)


aczechowski (approved):
Thanks @baardhermansen for the contribution! I did a heavy edit on the article, as it seems it hasn't been touched in over five years! 😮

FYI, for future reference, we don't like to use terms like "above" or "higher" for versions, which have an implicit value attached to them. Instead use terms like "later" or "earlier" that are purely time-based. 😄 [Style guide](https://docs.microsoft.com/style-guide/a-z-word-list-term-collections/l/later)

baardhermansen:
Thanks for the comment @aczechowski , i'll be sure to remember it. I simply used the language found on other Microsoft pages ;)

PR #10864, opened 2022-09-16
Update volume-activation-management-tool.md

baardhermansen:
## Why

Updated lacking info and missing markdown. 

## Changes

Changed bold section Important to markdown.
Updated the Important text with Windows 11.


aczechowski (approved):
Thanks @baardhermansen - another heavy edit 😉 

In general, if the specified version is no longer supported, then it's best to just generalize the statement to something like "any supported version". Unless there's some specific reason to call out a version that's not supported anymore.

PR #10865, opened 2022-09-16
Update activate-using-active-directory-based-activation-client.md

baardhermansen:
## Why

Ease of reading.

## Changes

Updated Applies to with bullet points, so that each product appears by itself.


aczechowski (approved):
Thanks @baardhermansen 😄

PR #10888, opened 2022-09-26
Add RegistryMapped

schenardie:
## Why

Add clarity to existing documentation

- Closes #[Issue Number]

## Changes
Mapping to Group Policy setting which results on the same as MDM config.


vinaypamnani-msft:
@schenardie I reviewed the CSP definition, and it doesn't map to any GP setting, so I'm closing this PR without merging it.

PR #10902, opened 2022-09-29
Update deployment-service-overview.md

everydayintech:
Feature updates for Windows 10 and later policy and therefore WufB DS also works with Azure AD registered devices.

## Why

Tests showed that Azure AD registered devices also use Feature updates for Windows 10 and later policy


JohanFreelancer9 (approved):

aczechowski:
Thanks @everydayintech for the contribution, but it does require Azure AD join, not just registration. Also @mestew has rewritten this article recently. I look forward to your next contribution to our content! 🙂

Issue #10862, opened 2022-09-16
loadstate example user mapping incorrect


teamneroblanco:
At the bottom there is a loadstate example, but it is mapping source user1 to destination user2, instead of what it says the intent is which is to migrate user1 and user2 from contoso to fabrikam, not migrate user1 to user2

[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 89c5c9e7-3e5e-b448-e3ce-f4f0ca5bfbc0
* Version Independent ID: bee454fd-4983-ab4d-b330-62a21cb61aef
* Content: [Migrate User Accounts (Windows 10) - Windows Deployment](https://docs.microsoft.com/en-us/windows/deployment/usmt/usmt-migrate-user-accounts)
* Content Source: [windows/deployment/usmt/usmt-migrate-user-accounts.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/deployment/usmt/usmt-migrate-user-accounts.md)
* Product: **w10**
* GitHub Login: @aczechowski
* Microsoft Alias: **aaroncz**

PR #10942, opened 2022-10-17
changedf user 2 to user1 and viceversa

RAJU2529:
as per user report #10862, so i changed  user2 to user 1 and viceversa

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/deployment/usmt/usmt-migrate-user-accounts.md
+++ b/windows/deployment/usmt/usmt-migrate-user-accounts.md
@@ -72,7 +72,7 @@ Links to detailed explanations of commands are available in the Related Topics s
 
 3.  Specify the following:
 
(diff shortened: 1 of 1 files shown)


JohanFreelancer9 (approved):

yogkumgit:
@frankroj Please review and signoff this PR. Thanks

yogkumgit:
@frankroj Please review and signoff this PR. Thanks

frankroj:
The change is most likely not correct. The correct syntax most likely is:

LoadState.exe  [\\server\share\migration\mystore](file://server/share/migration/mystore) /mu:contoso\user1:fabrikam\user1 /i:MigDocs.xml /i:MigApp.xml

However, I am confirming with experts. I think the main problem is the stated scenario - it talks about migrating two users but then only restores one user. I may just change the scenario so that both users are restored.

frankroj:
Fixed by https://github.com/MicrosoftDocs/windows-docs-pr/pull/7597

PR #10953, opened 2022-10-22
removed old version added new version adobe reader

RAJU2529:
after reading this article, i removed old version  and added a new version **2200320258** of adobe acrobat reader


JohanFreelancer9 (approved):

aczechowski (approved):
Thanks for the update

PR #11003, opened 2022-11-10
Update DORestrictPeerSelectionBy

GrischaE1:
We need to update the Policy to reflect the new options – 0 for NAT and 2 for Local Peer Discovery. See the DO reference article:
https://learn.microsoft.com/en-us/windows/deployment/do/waas-delivery-optimization-reference#select-a-method-to-restrict-peer-selection

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/client-management/mdm/policy-csp-deliveryoptimization.md
+++ b/windows/client-management/mdm/policy-csp-deliveryoptimization.md
@@ -1457,9 +1457,11 @@ ADMX Info:
 <!--/Scope-->
 <!--Description-->
 Set this policy to restrict peer selection via selected option.
-Options available are: 1=Subnet mask (more options will be added in a future release).
+In Windows 11 the 'Local Peer Discovery' option was introduced to restrict peer discovery to the local network. Currently the available options include: 0 = NAT, 1 = Subnet mask, and 2 = Local Peer Discovery. These options apply to both Download Modes LAN (1) and Group (2) and therefore means there is no peering between subnets. The default value in Windows 11 is set to "Local Peer Discovery".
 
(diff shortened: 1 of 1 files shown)


cmknox:
#SignOff

windows/client-management/mdm/policy-csp-deliveryoptimization.md
```diff
@@ -1457,9 +1457,11 @@ ADMX Info:
 <!--/Scope-->
 <!--Description-->
 Set this policy to restrict peer selection via selected option.
-Options available are: 1=Subnet mask (more options will be added in a future release).
+In Windows 11 the 'Local Peer Discovery' option was introduced to restrict peer discovery to the local network. Currently the available options include: 0 = NAT, 1 = Subnet mask, and 2 = Local Peer Discovery. These options apply to both Download Modes LAN (1) and Group (2) and therefore means there is no peering between subnets. The default value in Windows 11 is set to "Local Peer Discovery".
```
JohanFreelancer9, line 1460:
```suggestion
In Windows 11 the 'Local Peer Discovery' option was introduced to restrict peer discovery to the local network. Currently, the available options include: 0 = NAT, 1 = Subnet mask, and 2 = Local Peer Discovery. These options apply to both Download Modes LAN (1) and Group (2) and therefore it means that there is no peering between subnets. The default value in Windows 11 is set to "Local Peer Discovery".
```

JohanFreelancer9 (changes requested):
Copy review - @GrischaE1, kindly commit the minor suggested changes. Many thanks.

JohanFreelancer9 (approved):

PR #11100, opened 2022-12-19
Modify for convenience PIN

rui0122:
Adding "However" which is easier to follow explanation.

## Why

- Closes #[Issue Number]


aczechowski (approved):
Thanks @rui0122 for the edit, and thank you for your first contribution to Windows docs! 🎉

PR #11103, opened 2022-12-21
Clarification of Settings Catalog options

wattbt:
We currently list both a method of configuring this with Settings Catalog, and with a Custom profile. Arguably the custom profile should just go these days, but in any case the Settings Catalog entry was incomplete and needed clarification on the Telemetry options, as we did not state recommended settings. Those settings are also not required, but recommended, so I have moved them as such.

I've also added brief wording to clarify that you need not do a Settings Catalog AND a custom profile, as this has been misunderstood before.

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/deployment/update/wufb-reports-configuration-intune.md
+++ b/windows/deployment/update/wufb-reports-configuration-intune.md
@@ -8,7 +8,7 @@ author: mestew
 ms.author: mstewart
 ms.localizationpriority: medium
 ms.topic: article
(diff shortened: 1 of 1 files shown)


JohanFreelancer9 (approved):

aczechowski (approved):
Thanks @wattbt for the edit! I made one minor tweak to use the terminology "diagnostic data" as preferred (even though the settings may still use another term that we try to avoid 😉 )

PR #11123, opened 2022-12-23
added windows 11

RAJU2529:
after reading this article, i confirmed windows 11 is supported


JohanFreelancer9 (approved):

aczechowski:
I need to check on this one - my assumption is that this policy doesn't apply to client OS at all, but only applies to Windows Server.

aczechowski (approved):

PR #11258, opened 2023-01-05
Update hello-faq.yml

Ronald-den-Os:
Typo in line 83 (exiting should be existing)

## Why

- Closes #[Issue Number]


aczechowski (approved):
Thanks @Ronald-den-Os for catching that! 😄

PR #11263, opened 2023-01-09
Add missing info about RequirePrivateStoreOnly

jangrundmann:
## Description
ApplicationManagement/RequirePrivateStoreOnly is missing from the list. HoloLens release notes (https://learn.microsoft.com/en-us/hololens/hololens-release-notes#windows-holographic-version-21h2 and https://learn.microsoft.com/en-us/hololens/hololens-release-notes#use-only-private-store-apps-for-microsoft-store) clearly states it's available since 21H2

## Why
I noticed the HoloLens2 Policy CSP is missing an item that is mentioned elsewhere as available for this platform.

## Changes
added a new line with a link to the missing setting


aczechowski (approved):
Thanks @jangrundmann for the edit, and thank you for your first contribution to the windows docs! 🎉

PR #11266, opened 2023-01-10
Update understanding-admx-backed-policies.md

aendrawos:
The tool is supposed to encode and decode (it was written "encoding and encoding")

## Why

- Closes #[Issue Number]


aczechowski (approved):
Thanks @aendrawos for catching that and making the edit, and thank you for your first contribution to Windows docs! 🎉 

I also switched the URL to `https` since the site supports it. #SecureByDefault 😁

Issue #11265, opened 2023-01-10
incorrect cross reference


tilmanschmidt:
The document contains a cross reference for the topic of data recovery agents:

> For more information about adding data recovery agents, see [BitLocker basic deployment](https://learn.microsoft.com/en-us/windows/security/information-protection/bitlocker/bitlocker-basic-deployment).

But on the referenced page the term "data recovery agent" isn't mentioned anywhere.

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 5476099c-9d70-97df-8ee6-bc442605cfbd
* Version Independent ID: 49ca0628-bd88-e05b-b705-3b361b57512d
* Content: [BitLocker Group Policy settings (Windows 10)](https://learn.microsoft.com/en-us/windows/security/information-protection/bitlocker/bitlocker-group-policy-settings#choose-how-bitlocker-protected-operating-system-drives-can-be-recovered)
* Content Source: [windows/security/information-protection/bitlocker/bitlocker-group-policy-settings.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/information-protection/bitlocker/bitlocker-group-policy-settings.md)
* Product: **windows-client**
* Technology: **itpro-security**
* GitHub Login: @frankroj
* Microsoft Alias: **frankroj**

msbemba:
@tilmanschmidt Thank you for your feedback. We have removed the incorrect reference via PR#https://github.com/MicrosoftDocs/windows-itpro-docs/pull/11290

Once the author approves, the changes will be live.

Also, once another document is made available, the correct reference will be added.

Thanks
Sri

frankroj:
The reference is actually correct. "Data recovery agents" is just a generic term that can refer to any method to recover the key which can then be used to recover data on the drive. The linking article goes into detail regarding the different data recovery agents, for example:

The recovery key can be stored using the following methods:

Save to your Azure AD account (if applicable)
Save to a USB flash drive
Save to a file - the file needs to be saved to a location that isn't on the computer itself such as a network folder or OneDrive
Print the recovery key

Closing issue since there is nothing to fix.

tilmanschmidt:
May I suggest adding your clarifying comment to the target page [BitLocker basic deployment](https://learn.microsoft.com/en-us/windows/security/information-protection/bitlocker/bitlocker-basic-deployment) then?
[...]

PR #11290, opened 2023-01-24
Update bitlocker-group-policy-settings.md

msbemba:
Remove the BitLocker guide as the reference was incorrect. No reference document to link about data recovery agents

Fixes#https://github.com/MicrosoftDocs/windows-itpro-docs/issues/11265


msbemba:
@JohanFreelancer9 Please copy review

JohanFreelancer9 (approved):

yogkumgit:
@frankroj Please merge this PR. Thanks

frankroj:
The reference is actually correct. "Data recovery agents" is just a generic term that can refer to any method to recover the key which can then be used to recover data on the drive. The linking article goes into detail regarding the different data recovery agents, for example:

The recovery key can be stored using the following methods:

Save to your Azure AD account (if applicable)
Save to a USB flash drive
Save to a file - the file needs to be saved to a location that isn't on the computer itself such as a network folder or OneDrive
Print the recovery key

Closing this PR without merging. I will also close out the issue.

PR #11297, opened 2023-01-25
Correct mistake in event-4670.md

ep3p:
## Description

Fill missing information of ACE strings according to:
- https://learn.microsoft.com/en-us/windows/win32/secauthz/ace-strings
- https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/f4296d69-1c0f-491f-9587-a960b292d070?redirectedfrom=MSDN

And move other information to other column.

## Why

Because the information is missing, or the information is not in the correct column.

## Changes

It just modifies some strings in a table.


vinaypamnani-msft:
Thanks @ep3p for the contribution! I've merged these changes :)

PR #11342, opened 2023-02-13
Clarification on GPO effect on restrictedAdmin

martinsohn:
## Why

My testing shows that Restricted Admin mode cannot be enforced with "mstsc.exe /remoteAdmin" when "Restrict Credential Delegation" is enabled. I had previously assumed this but it seems not to be the case. A clarification would be useful for others.

## Changes

Clarifying the effect of "Restrict Credential Delegation" on "mstsc.exe /remoteAdmin".

### DIFF
--- a/windows/security/identity-protection/remote-credential-guard.md
+++ b/windows/security/identity-protection/remote-credential-guard.md
@@ -156,6 +156,7 @@ Beginning with Windows 10 version 1703, you can enable Windows Defender Remote C
 
      > [!NOTE]
      > Neither Windows Defender Remote Credential Guard nor Restricted Admin mode will send credentials in clear text to the Remote Desktop server.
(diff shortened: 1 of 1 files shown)


windows/security/identity-protection/remote-credential-guard.md
```diff
@@ -156,6 +156,7 @@ Beginning with Windows 10 version 1703, you can enable Windows Defender Remote C
 
      > [!NOTE]
      > Neither Windows Defender Remote Credential Guard nor Restricted Admin mode will send credentials in clear text to the Remote Desktop server.
+     > When **Restrict Credential Delegation** is enabled the /restrictedAdmin switch has no effect; consequently, Windows Defender Remote Credential Guard will be preferred.
```
mapalko, line 159:
```suggestion
     > When **Restrict Credential Delegation** is enabled, the /restrictedAdmin switch will be ignored. Windows will enforce the policy configuration instead and will use Windows Defender Remote Credential Guard. 
```

paolomatarazzo:
Thank you @martinsohn for the contribution!

PR #11346, opened 2023-02-15
Spelling error in the "Policy timeline" section

nx9010:
There is a small spelling error. In the "Policy timeline" section there is a sentence written like this: "accessgroup dec" while the correct one would be "accessgroup desc".

## Why

- Closes #[Issue Number]


JohanFreelancer9 (approved):

vinaypamnani-msft:
Thanks for the contribution @nx9010! I've merged the change.

PR #11413, opened 2023-03-17
Update connect-to-remote-aadj-pc.md

SAINT-SCOBBERLOTCHER:
I added some new suggested verbiage to the page.

Also some other feedback. 
- It took me a while to figure out that the Supported configurations table only applied to the Connect without Azure AD authentication section.  Can you add a table or note to the Connect with Azure AD Authentication section that either enumerates all the credential types that are supported or that makes this more clear?  To me it first made me think that credentials like FIDO2 security keys were not supported with Azure AD Authentication, but after testing I discovered that they are in fact supported.  

- Can you add notes about how long the RDP session lives by default. WIth Azure AD Authentication, even signing out of the remote desktop does not cause the user to reauth again when signing back in. The session length seems quite long. Maybe add a suggestion for CA Policy Session Controls to limit the session length also. 

- The App name/appID needed in the CA Policy isn't easy to find, so I put a specific mention of it in the page.

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/client-management/connect-to-remote-aadj-pc.md
+++ b/windows/client-management/connect-to-remote-aadj-pc.md
@@ -49,13 +49,14 @@ There's no requirement for the local device to be joined to a domain or Azure AD
 To connect to the remote computer:
 
 - Launch **Remote Desktop Connection** from Windows Search, or by running `mstsc.exe`.
-- Specify the name of the remote computer.
+- Specify the **name** of the remote computer (IP address cannot be used when **Use a web account to sign in to the remote computer** option is used.) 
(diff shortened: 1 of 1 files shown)


vinaypamnani-msft:
Thanks for the first-time contribution @SAINT-SCOBBERLOTCHER! I've made minor tweaks to the changes you suggested, and am merging this PR. 

- I removed the specific CA controls, and linked to the CA documentation instead.
- Expanded a statement to clarify the supported configuration applicability. I'll investigate more for the Azure AD authentication and add a table there when I have more info,

I didn't add anything about the session length, as it's not RDP specific, but I'll investigate more on this and update later if needed. Looking forward to more contributions from you.

SAINT-SCOBBERLOTCHER:
Thanks @vinaypamnani-msft , I'm impressed by the speed this was completed. Thank you so much for considering the edits. I look forward to contributing more in the future.

PR #11416, opened 2023-03-20
Update hello-faq.yml

mudeeb:
PIN 1231 doesn't have a constant delta (1,1,8), so it's allowed ====>must be change to ====>PIN 1231 doesn't have a constant delta (1,1,2), so it's allowed.

8 to 2 this is the change

## Why

- Closes #[Issue Number]


paolomatarazzo:
Thank you for catching it, @mudeeb

Issue #11453, opened 2023-04-12
Need to provide an example of a CSP that does not apply for MDMWinsOverGP


YongRhee-MSFT:
🚨 Looks like you arrived from the internal review site 
🚨 Do not enter Microsoft confidential information here

[Enter feedback here]

Need to provide an example of a CSP that does not apply for MDMWinsOverGP

Current:
 It does not apply to other MDM settings with equivalent GP settings that are defined in other CSPs.

Recommendation:
 It does not apply to other MDM settings with equivalent GP settings that are defined in other CSPs such as [Defender CSP](https://learn.microsoft.com/en-us/windows/client-management/mdm/defender-csp).
---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: becd9bd1-6396-3f4a-cc78-c9a06b781516
* Version Independent ID: efeec2a0-08b0-d1fc-72c5-14c23fa11397
* Content: [ControlPolicyConflict Policy CSP - Windows Client Management](https://review.learn.microsoft.com/en-us/windows/client-management/mdm/policy-csp-controlpolicyconflict)
* Content Source: [windows/client-management/mdm/policy-csp-controlpolicyconflict.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/client-management/mdm/policy-csp-controlpolicyconflict.md)
* Product: **windows-client**
* Technology: **itpro-manage**
* GitHub Login: @vinaypamnani-msft
* Microsoft Alias: **vinpa**

YongRhee-MSFT:
Thx Vinay.

PR #11454, opened 2023-04-12
Update defender-csp.md

YongRhee-MSFT:
Adding info about MDMWinsOverGP not being relevant for the Defender CSP.  MDMWinsOverGP only applies to the Windows Client Management CSP.

## Why

- Closes https://github.com/MicrosoftDocs/windows-itpro-docs/issues/11453


vinaypamnani-msft:
@YongRhee-MSFT thanks for the contribution! I've moved the note you added inside an editable section so it's not overwritten by automation.

Issue #11435, opened 2023-04-03
[Account Lockout Policy] Group Policy object location is incorrect


technicallynotdeaf:
Hi Team,

In the article about "[account lockout policy](https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/account-lockout-duration)" settings, lots of good info is given about settings which control incorrect password attempt lockout thresholds, duration etc.

Under the heading "Location", the location of the group policy object is **incorrectly** specified as 
Computer Configuration\Windows Settings\Security Settings\Account Policies\**Password Policy**

The incorrect location ('password policy') contains settings related to length, complexity requirements etc - and not the ones addressed in the article.

The correct GPO object location should be 
Computer Configuration\Windows Settings\Security Settings\Account Policies\**Account Lockout Policy**

Thanks and have a great day :) 

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 403fdccd-6988-3a5d-d175-c8f30ffddf31
* Version Independent ID: 5fd127aa-d224-559a-c7ad-00a1257397b7
* Content: [Account lockout duration (Windows 10)](https://learn.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/account-lockout-duration)
* Content Source: [windows/security/threat-protection/security-policy-settings/account-lockout-duration.md](https://github.com/MicrosoftDocs/windows-itpro-docs/blob/public/windows/security/threat-protection/security-policy-settings/account-lockout-duration.md)
* Product: **windows-client**
* Technology: **itpro-security**
* GitHub Login: @vinaypamnani-msft
* Microsoft Alias: **vinpa**

RAJU2529:
I will check today and edit.

PR #11455, opened 2023-04-12
reanmed Password Policy

RAJU2529:
as per user report #11435 , i changed  from **Password Policy**  to   **Account Lockout Policy**

Screenshot 2023-04-12 194905


vinaypamnani-msft:
Thanks @RAJU2529! Merging...

PR #11526, opened 2023-05-22
Update note about using atomic blocks with firewall CSP

CatalinFetoiu:
Update note about using atomic blocks with firewall CSP

### DIFF
--- a/windows/client-management/mdm/firewall-csp.md
+++ b/windows/client-management/mdm/firewall-csp.md
@@ -22,7 +22,9 @@ The Firewall configuration service provider (CSP) allows the mobile device manag
 
 > [!NOTE]
 > Firewall rules in the FirewallRules section must be wrapped in an Atomic block in SyncML, either individually or collectively.
+>
 > Atomic blocks are "all or nothing", if a firewall rule or firewall setting in an Atomic block fails to be applied, the entire Atomic block will fail to be applied.
+>
 > If an Atomic block contains a firewall rule or firewall setting that is not supported on a particular Windows OS version, the entire Atomic block will fail to be applied on that Windows version. For example, firewall rules with IcmpTypesAndCodes are only supported on Windows 11, applying an Atomic block that contains a rule with IcmpTypesAndCodes on Windows 10 will fail.
 
 For detailed information on some of the fields below, see [[MS-FASP]: Firewall and Advanced Security Protocol documentation](/openspecs/windows_protocols/ms-winerrata/6521c5c4-1f76-4003-9ade-5cccfc27c8ac).
--- a/windows/client-management/mdm/firewall-csp.md
+++ b/windows/client-management/mdm/firewall-csp.md
@@ -22,6 +22,8 @@ The Firewall configuration service provider (CSP) allows the mobile device manag
 
 > [!NOTE]
 > Firewall rules in the FirewallRules section must be wrapped in an Atomic block in SyncML, either individually or collectively.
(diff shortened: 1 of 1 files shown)


windows/client-management/mdm/firewall-csp.md
```diff
@@ -22,6 +22,8 @@ The Firewall configuration service provider (CSP) allows the mobile device manag
 
 > [!NOTE]
 > Firewall rules in the FirewallRules section must be wrapped in an Atomic block in SyncML, either individually or collectively.
+> Atomic blocks are "all or nothing", if a firewall rule or firewall setting in an Atomic block fails to be applied, the entire Atomic block will fail to be applied.
+> If an Atomic block contains a firewall rule or firewall setting that is not supported on a particular Windows OS version, the entire Atomic block will fail to be applied on that Windows version. For example, firewall rules with IcmpTypesAndCodes are only supported on Windows 11, applying an Atomic block that contains a rule with IcmpTypesAndCodes on Windows 10 will fail.
```
JohanFreelancer9, line 26:
```suggestion
>
> Atomic blocks are "all or nothing", if a firewall rule or firewall setting in an Atomic block fails to be applied, the entire Atomic block will fail to be applied.
>
> If an Atomic block contains a firewall rule or firewall setting that is not supported on a particular Windows OS version, the entire Atomic block will fail to be applied on that Windows version. For example, firewall rules with IcmpTypesAndCodes are only supported on Windows 11, applying an Atomic block that contains a rule with IcmpTypesAndCodes on Windows 10 will fail.
```

JohanFreelancer9 (changes requested):
Copy review - @CatalinFetoiu, kindly commit the suggested change. Many thanks.

windows/client-management/mdm/firewall-csp.md
```diff
@@ -22,6 +22,10 @@ The Firewall configuration service provider (CSP) allows the mobile device manag
 
 > [!NOTE]
 > Firewall rules in the FirewallRules section must be wrapped in an Atomic block in SyncML, either individually or collectively.
+>
+> Atomic blocks are "all or nothing", if a firewall rule or firewall setting in an Atomic block fails to be applied, the entire Atomic block will fail to be applied.
```
vinaypamnani-msft, line 26:
```suggestion
> Atomic blocks are "all or nothing", if a firewall rule or firewall setting in an Atomic block fails to be applied, the entire Atomic block fails to be applied.
```

windows/client-management/mdm/firewall-csp.md
```diff
@@ -22,6 +22,10 @@ The Firewall configuration service provider (CSP) allows the mobile device manag
...
+>
+> If an Atomic block contains a firewall rule or firewall setting that is not supported on a particular Windows OS version, the entire Atomic block will fail to be applied on that Windows version. For example, firewall rules with IcmpTypesAndCodes are only supported on Windows 11, applying an Atomic block that contains a rule with IcmpTypesAndCodes on Windows 10 will fail.
```
vinaypamnani-msft, line 28:
```suggestion
> If an Atomic block contains a firewall rule or firewall setting that is not supported on a particular Windows OS version, the entire Atomic block fails to be applied on that Windows version. For example, firewall rules with IcmpTypesAndCodes are only supported on Windows 11, applying an Atomic block that contains a rule with IcmpTypesAndCodes on Windows 10 fails.
```

vinaypamnani-msft (approved):

PR #11532, opened 2023-05-24
Update personalization-csp.md

docupdates:
Adds note about WinPro compatibility with desktop image url

## Why
DesktopImageURL reports as not compatible in Intune for Windows 10 Pro, after discussion with support it seems there's a relevant internal note about it saying it's not compatible with Windows 10 Pro. This is supported by the UI text in Intune excluding Pro:
image

The Intune docs point to this being the CSP that backs the setting above: https://learn.microsoft.com/en-us/mem/intune/configuration/device-restrictions-windows-10#personalization

I would simply have suggested it be listed as incompatible, but have found contradictory information from an MS employee here ([Lu Dai-MSFT](https://learn.microsoft.com/en-us/users/na/?userid=c9083ec3-b3f6-48ad-a978-5924127a5d72)):

[Windows 10 MDM - Personalization-csp, DesktopImageUrl setting failing.](https://learn.microsoft.com/en-us/answers/questions/622758/windows-10-mdm-personalization-csp-desktopimageurl)

They state the following

_It works in Windows 10 Pro and Windows 10 Pro in S mode if SetEduPolicies in SharedPC CSP is set._ 

I'm not in a position to test that however - but a simple YES on compatibility here isn't quite accurate. I might be able to convince support to give me the internal article ID to pass it on to you if you need it for validation on the compatibility.

## Changes

- Adds compatibility note for DesktopImageURL regarding Windows Pro compatibility.
-

### DIFF
--- a/windows/client-management/mdm/personalization-csp.md
+++ b/windows/client-management/mdm/personalization-csp.md
@@ -19,6 +19,9 @@ ms.topic: reference
 <!-- Personalization-Editable-Begin -->
 <!-- Add any additional information about this policy here. Anything outside this section will get overwritten. -->
 The Personalization CSP can set the lock screen and desktop background images. Setting these policies also prevents the user from changing the image. You can also use the Personalization settings in a provisioning package.
+
+> [!IMPORTANT]
+> Personalization CSP is supported in Windows Enterprise and Education SKUs. It works in Windows Professional only when SetEduPolicies in [SharedPC CSP](sharedpc-csp.md) is set.
 <!-- Personalization-Editable-End -->
 
 <!-- Personalization-Tree-Begin -->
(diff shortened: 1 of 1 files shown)


windows/client-management/mdm/personalization-csp.md
```diff
@@ -88,6 +88,12 @@ This represents the status of the DesktopImage. 1 - Successfully downloaded or c
 <!-- Device-DesktopImageUrl-Description-Begin -->
 <!-- Description-Source-DDF -->
 A http or https Url to a jpg, jpeg or png image that needs to be downloaded and used as the Desktop Image or a file Url to a local image on the file system that needs to be used as the Desktop Image.
+
+> [!NOTE]
+> This setting is only compatible with Windows 10 Pro if SetEduPolicy is enabled in 
+[SharedPC](sharedpc-csp.md)
+>
+> This only applies to Windows 10 Pro and not the other editions listed
```
JohanFreelancer9, line 96:
```suggestion
> This setting is only compatible with Windows 10 Pro if SetEduPolicy is enabled in 
[SharedPC](sharedpc-csp.md).
>
> This only applies to Windows 10 Pro and not the other editions listed.
```

JohanFreelancer9 (changes requested):
Copy review - @docupdates, kindly commit the suggested minor change. Many thanks.

docupdates:
> Copy review - @docupdates, kindly commit the suggested minor change. Many thanks.

Thanks for the review! Changes committed!

docupdates:
> @docupdates This article used to have this note for the entire CSP, but I recently automated the CSP articles, and the note was lost in translation. I've added it back, and removed your edit.

Even better, thank you!

PR #11540, opened 2023-05-29
Update microsoft-recommended-block-rules.md

brunocarranca:
Removed Intune Agent from xml since blocking it get the machine into unsupported scenario regarding Intune

## Why

- Closes #[Issue Number]


JohanFreelancer9 (approved):

jsuther1974:
The block is version specific, though, and only targets an old version of Intune. What's the scenario where this is actually causing a functional break to Intune?

aczechowski:
@brunocarranca can you please clarify your intent with this edit, see Jeffrey's question:

> The block is version specific, though, and only targets an old version of Intune. What's the scenario where this is actually causing a functional break to Intune?

aczechowski:
Sent internal email to followup

vinaypamnani-msft:
Closing this PR per internal thread.

PR #11548, opened 2023-06-02

miguel-s-ferreira:
Update link to the specific application ID related with Windows Store for Business, as they share the same AppID

## Why

- Closes #[Issue Number]


frankroj:
Thank you for the contribution @miguel-s-ferreira. However I am going to reject these edits for the following two reasons:

1) You bolded within the link name. Bolding doesn't work within link names, nor is it needed since link names are already bolded.
2) The text search that you put as part of the URL searches the whole article and not just the specified section. If in the future, Windows Store for Business was added somewhere else in the article before this section, it would jump to the wrong location. For this reason, we only stick with URLs that jump to sections (driven by unique headers) and not to text searches sinc text searches may not be unique.

If you feel there is another way you might be able to get your point across using different type of edits, then go ahead and submit and I will review.

PR #11570, opened 2023-06-13
Update hello-hybrid-cloud-kerberos-trust.md

zjorz:
I have 2 comments about notes half-way and at the end. Please consider in rewriting that as it is either not accurate, or dangerous

## Why

- Closes #[Issue Number]


zjorz:
the PR is just to get the attention. Please see the remarks

PR #11572, opened 2023-06-14
Clarified content around WDS as it relates to MDT.

Andrew-Nielsen:
## Description
The document makes the assumption the MDT still supports Windows 11. Referenced another document to indicate it does not to not confuse customers on utilizing MDT when they shouldn't.

## Why
This pull request clarifies that customers don't have a good option to utilize MDT with Windows 11.

## Changes
Added a sentence and provided another link to the document where MDT mentions it isn't supported.

### DIFF
--- a/windows/deployment/wds-boot-support.md
+++ b/windows/deployment/wds-boot-support.md
@@ -38,7 +38,7 @@ The table below provides support details for specific deployment scenarios. Boot
 
 ## Reason for the change
 
-Alternatives to WDS, such as [Microsoft Configuration Manager](/mem/configmgr/) and [Microsoft Deployment Toolkit](/mem/configmgr/mdt/) (MDT) provide a better, more flexible, and feature-rich experience for deploying Windows images.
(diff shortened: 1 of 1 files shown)


frankroj:
Thank you for the feedback and contribution @Andrew-Nielsen. I've reformatted and reworded the contribution to fit our style guidelines, but the same message is still delivered. I also changed the link from known issues to supported platforms. MDT not supporting Windows 11 isn't a known issue - its expected behavior as documented under Supported platforms. This should publish either later today or sometime tomorrow.

Andrew-Nielsen:
No problem. Interesting @frankroj. Gotcha. The Note of "MDT is not supported with Windows 11." is listed under the known issues document. That's why I referenced that. Maybe it was included to list up to the supported platforms document.

frankroj:
Yes, its listed there because it goes on to discuss some issues when trying to use Windows 11 with MDT, but even the Important note at the top has a link back to the Supported platforms doc. Specifically, the scenarios covered are indirect issues, such as MDT integration with ConfigMgr (which does support Windows 11) and trying to use the Windows 11 ADK with MDT.

PR #11573, opened 2023-06-14
Update policy-csp-defender.md

HarveyHuangMSFT:
Hi Team, there needs to be some slight corrections with the DisableCatchupFullScan section. If we set it to not configured or enabled, catch-up full scans will occur. If we set it to disabled. catch-up full scans will not occur. This is to keep it consistent with this article here: https://learn.microsoft.com/en-us/powershell/module/defender/set-mppreference?view=windowsserver2022-ps

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/client-management/mdm/policy-csp-defender.md
+++ b/windows/client-management/mdm/policy-csp-defender.md
@@ -1544,9 +1544,9 @@ This policy setting defines the number of days items should be kept in the Quara
 <!-- Description-Source-ADMX -->
 This policy setting allows you to configure catch-up scans for scheduled full scans. A catch-up scan is a scan that's initiated because a regularly scheduled scan was missed. Usually these scheduled scans are missed because the computer was turned off at the scheduled time.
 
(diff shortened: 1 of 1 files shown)


JohanFreelancer9 (approved):

aczechowski:
Hi @HarveyHuangMSFT thanks for the edit, but the strings in that section come directly from the ADMX, you can see they are the same as what's in the group policy:

image

This content is now mostly automated, so only edits to the "editable" sections are preserved from month to month, otherwise it just uses what's in the DDF and ADMX XML. So even if we accept this change it will revert next month.

Since there's a discrepancy between what's shown here and in group policy, and what's in PowerShell, it's probably best to file feedback via the Feedback Hub so that the appropriate engineering team can sort out which is correct.

For more info, see [Send feedback to Microsoft with the Feedback Hub app](https://support.microsoft.com/windows/send-feedback-to-microsoft-with-the-feedback-hub-app-f59187f8-8739-22d6-ba93-f66612949332). I think it’s important that you submit it directly, because that allows you to receive notifications and more closely monitor the progress. I'm closing this PR.

PR #11578, opened 2023-06-21
Update policy-csp-userrights.md

jukishiya:
The mention of "DenyLocalLogOn" is incorrect, so I will correct it to the accurate information. (Currently, the content of "DenyLogOnAsService" is mistakenly mentioned in this section.)

### DIFF
--- a/windows/client-management/mdm/policy-csp-userrights.md
+++ b/windows/client-management/mdm/policy-csp-userrights.md
@@ -894,10 +894,10 @@ This user right determines which users are prevented from accessing a computer o
 
 <!-- DenyLocalLogOn-Description-Begin -->
 <!-- Description-Source-DDF -->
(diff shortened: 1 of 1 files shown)


jukishiya:
#sign-off

aczechowski:
Hi @jukishiya, thanks for the contribution, but this content is now mostly autogenerated from the DDF and ADMX XML. Only the sections marked "editable" are preserved from month to month, so if we were to accept this change, it will be overwritten next month. So there are two ways to address this:

1. Close this PR and file feedback and/or bug to engineering to fix the source string.
2. Move your change as a note into the editable section.

Let me know how you want to proceed.

aczechowski (changes requested):
See prior comment - can't accept as-is

PR #11581, opened 2023-06-22
Patch 2

chcart:
## Description

Minor Note addition to TLS 1.3 text
## Why

Customers and Microsoft employees think that experimental technologies are supported.


aczechowski (approved):
Thanks @chcart for the contribution! This article is for a version of Windows 10 that's no longer supported, and shouldn't be easily discoverable (shouldn't be indexed by search engines), but we maintain for historical purposes. But there's no harm in adding the note. I made a small tweak for style, but otherwise looks good.

PR #11585, opened 2023-06-28
Adding clarification about mode 100 - not recommended

narkissit:
adding clarification that Download Mode 100 is not recommended.

## Why

- Closes #[Issue Number]


JohanFreelancer9 (approved):

aczechowski (approved):
Thanks @narkissit for the edit, I made a few tweaks but otherwise should be good.

PR #11586, opened 2023-06-28
clarifying that bypass is not recommended.

narkissit:
## Why

- Closes #[Issue Number]


JohanFreelancer9 (approved):

aczechowski (approved):
Thanks @narkissit for the edit, I made a few tweaks but otherwise should be good.

PR #11593, opened 2023-07-03
Update interactive-logon-dont-display-username-at-sign-in.md

ryanriesMSFT:
Some needed clarification as it is not clear how this policy is not compatible with certain other logon workflows.

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/security/threat-protection/security-policy-settings/interactive-logon-dont-display-username-at-sign-in.md
+++ b/windows/security/threat-protection/security-policy-settings/interactive-logon-dont-display-username-at-sign-in.md
@@ -89,7 +89,7 @@ Enable the **Interactive logon: Don't display user name at sign-in** setting.
 
 ### Potential impact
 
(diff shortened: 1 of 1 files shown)


windows/security/threat-protection/security-policy-settings/interactive-logon-dont-display-username-at-sign-in.md
```diff
@@ -89,7 +89,7 @@ Enable the **Interactive logon: Don't display user name at sign-in** setting.
 
 ### Potential impact
 
-Users must always type their usernames and passwords when they log on locally or to the domain. The sign in tiles of all logged on users aren't displayed.
+Users must always type their usernames and passwords when they log on locally or to the domain. The sign in tiles of all logged on users aren't displayed. When this policy is Enabled, you will be unable to change the default credential provider to anything other than username/password. In addition, this policy may be incompatible with Autologon and Multi-Factor Unlock.
```
vinaypamnani-msft, line 92:
```suggestion
Users must always type their usernames and passwords when they log on locally or to the domain. The sign in tiles of all logged on users aren't displayed. When this policy is enabled, you will be unable to change the default credential provider to anything other than username/password. In addition, this policy may be incompatible with autologon and multi-factor unlock.
```

PR #11599, opened 2023-07-04
Windows Drivers update setting to block drivers

silvermarkg:
In the 'Locations of policies that exclude drivers' list, the Windows Drivers update setting indicates being set to Allow blocks drivers. I believe this should be **Block** and not **Allow**.


JohanFreelancer9 (approved):

Ahmedgama12345 (changes requested):
ارجو اغلاق جميع الاوامر والحزم وحزرهم من الوصول إلى حسابي من تاني

mestew:
verified update ring setting, merging this

Lo236816:
Revert

PR #11626, opened 2023-07-24
Use the "/genconfig option" instead of "use the ScanState Syntax option"

Urbs3w:
## Why

The part that said, "use the ScanState Syntax option" didn't make sense. In the next part, it seemed more helpful to link directly to the relevant section in the linked article than linking to the top of the article. 

## Changes

The part that said, "use the ScanState Syntax option" didn't make sense, so--as you can probably see in the diff--I replaced that part with "use the /genconfig option" to make sense.

I also changed the link that's in the next part/next step to link directly to the "Migration rule options" section of the "ScanState syntax" article instead of the top of that article because that seems more helpful than linking to the top of the article. However, if it's Microsoft's style to link to the top of an article instead of directly to a relevant section, feel free to revert that change in order to be consistent with your prescribed style.


JohanFreelancer9 (approved):

Urbs3w:

frankroj:
Thank you for the contribution @Urbs3w. I made some changes to your contributions to further clarify meaning and to also account for style and grammar. The changes should go live in the next day or two.

PR #11710, opened 2023-09-15
Fix capitalization for RetryTimeInHours profilexml example

mitschmi:
## Why
Xml Node names are case sensitive, so this example does not actually work and fails to configure RetryTimeInHours.

## Changes
Fix capitalization in example


aczechowski (approved):
Thanks @mitschmi for the edit! I confirmed the change is in the CspMoreInfo editable section, so it should persist next month's automation :)

PR #11717, opened 2023-09-20
Fix spelling mistake: 'Allow system to be shut down without having to lg on'

jkrejcha:
## Why

"log on" is spelled incorrectly

## Changes

* This fixes the spelling of `log on` to its correct spelling :)


aczechowski (approved):
Thanks @jkrejcha for catching that, and submitting the edit!

PR #11721, opened 2023-09-22
update steps to enable nested virtualization

yunxiali:
The wording here is not clear on whether the command should be ran on host or guest. Also depending on the VM version a update is needed. So changed the wording and added the VM update step.


aczechowski (approved):
Thanks @yunxiali for the useful edit, and thank you for your first contribution to Windows client docs! 🎉 I made a small tweak to the wording for style.

PR #11746, opened 2023-10-09
Update policy-csp-defender.md

j0eyv:
The current documentation is confusing and wrong. Made it a bit more clear. See current documentation in this picture. Makes more sense to modify this..

https://pbs.twimg.com/media/F7_Q5p4WEAEaKAj?format=png&name=900x900

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/client-management/mdm/policy-csp-defender.md
+++ b/windows/client-management/mdm/policy-csp-defender.md
@@ -834,7 +834,7 @@ Allows or disallows Windows Defender Script Scanning functionality.
 <!-- Description-Source-ADMX -->
 This policy setting allows you to configure whether or not to display AM UI to the users.
 
-If you enable this setting AM UI won't be available to users.
+If you enable this setting AM UI will be available to users.
 <!-- AllowUserUIAccess-Description-End -->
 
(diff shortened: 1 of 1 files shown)


JohanFreelancer9 (approved):

aczechowski:
Thanks @j0eyv for submitting the edit, however most of the CSP documentation is now automated, and that specific string comes from the associated ADMX file. For more information, see [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference). When there's a new release of ADMX files, then it's possible that strings like these will be updated. But if we were to accept this edit, then it will be overwritten with the original string from the ADMX in a few weeks when the automation reruns. The best way to attempt to address this issue is to [send feedback to the Windows engineering team with the Feedback Hub app](https://support.microsoft.com/windows/send-feedback-to-microsoft-with-the-feedback-hub-app-f59187f8-8739-22d6-ba93-f66612949332). I think it’s important that you submit it directly, because that allows you to receive notifications and more closely monitor the progress. Make sure that you note the change is needed in the ADMX file for that policy.

PR #11794, opened 2023-11-02

v-aweber:
Changed enterprise app name Universal Store Service APIs and Web Application, to Windows Store for Business to reflect name changes.

<!-- 
## Description
Universal Store Service APIs and Web Application was renamed to Windows Store for Business 
## Why

Searching Conditional Access for Universal Store Service APIs and Web Application yields no results. Instead, you need to use the new name Windows Store for Business

### DIFF
--- a/windows/deployment/windows-10-subscription-activation.md
+++ b/windows/deployment/windows-10-subscription-activation.md
@@ -39,7 +39,7 @@ This article covers the following information:
 For more information on how to deploy Enterprise licenses, see [Deploy Windows Enterprise licenses](deploy-enterprise-licenses.md).
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


aczechowski:
Thanks @v-aweber - I'm no expert on this topic, but I did some hunting, and looking in one Azure tenant, the name there is "Universal Store Service APIs and Web Application":
image
So I think the doc is accurate. Given the retirement of "Store for Business" I'm assuming we shouldn't be using that term anymore. So I'm inclined to close this PR and leave the doc as-is...but let me know if you have other thoughts/justification for the change. Or if I need to dig deeper with others.

JohanFreelancer9 (approved):

v-aweber:
image

v-aweber:
It's possible that this varies by a tenant's creation date, but in many of my customers tenants this app is only searchable in conditional access as Windows Store for Business. I'll try to provide examples

v-aweber:
I had multiple peers confirm this behavior in conditional access 
image
image

v-aweber:
This is likely because of how the rename is only reflected in the home tenant
https://learn.microsoft.com/en-us/entra/identity-platform/app-objects-and-service-principals?tabs=browser#consequences-of-modifying-and-deleting-applications

v-aweber:
I'm going to go ahead and close this and open an issue as I'm discovering this is a more complex issue.

PR #11827, opened 2023-11-20
Correction: Change 23H3 to 23H2

lzw29107:
Description

The page title is displayed incorrectly as 23H3.

Changes
Correction: Change 23H3 to 23H2


DHB-MSFT:
@lzw29107  Thanks for catching that typo and submitting the fix.

PR #11859, opened 2023-12-27
Update design-create-appid-tagging-policies.md

marcushms:
Adding the implications of using the templates. Using the templates will tag all Windows/Microsoft binaries as well, which might not be intended.

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/security/application-security/application-control/windows-defender-application-control/AppIdTagging/design-create-appid-tagging-policies.md
+++ b/windows/security/application-security/application-control/windows-defender-application-control/AppIdTagging/design-create-appid-tagging-policies.md
@@ -22,7 +22,7 @@ You can use the Windows Defender Application Control (WDAC) Wizard and the Power
 	![Configuring the policy base and template.](../images/appid-wdac-wizard-1.png)
 	
 	> [!NOTE]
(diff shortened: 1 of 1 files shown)


jgeurten:
Personally, I think the implication of this is (hopefully) evident and need not be called out explicitly.

marcushms:
Hi Jordan!

Yes, personally I do think the same.
But we do have a customer that sees this a “little” different and insists on doc change. ☹
Hence why I made the change request today.

Thanks
Marcus

jgeurten:
Understood. Please set expectations with the customer that the doc will not be updated.

marcushms:
Okay, I will inform the engineer that was working with the customer.
Shouldn't we remove that note completely btw?
As far as I understood the performance issue was fixed two or three months ago, right?

Thanks
Marcus

aczechowski:
Thanks @marcushms for making the suggestion, but per @jgeurten's comment, I'm closing this PR. Even though this PR is closed, ya'll are welcome to continue the conversation here. Just realize it's public on GitHub. You could switch to Microsoft internal email/IM as well. 😉 

@marcushms FYI in the future I suggest using the private repo. That's generally best found by adding "review." to the start of the URL before you select the edit action. Feel free to ping me if you have questions.

PR #11866, opened 2024-01-26
Update operations-guide.md

lizhiquan:
Correct add BitLocker key protector command.


paolomatarazzo:
Thank you for catching it, @lizhiquan.

PR #11873, opened 2024-02-22
Update docs for Renew nodes in CertificateStore CSP docs.

kevinhwang:


kevinhwang:

vinaypamnani-msft:
Thanks for creating this PR @kevinhwang. The CSP docs are automated using DDF files from windows source, and the node description and format is in the DDF definition and any manual changes to these items will get overwritten when the automation runs. For this issue, please submit feedback via Feedback Hub app. Once the feedback is addressed, the change will flow into the DDF files, and docs will get automatically updated. For more info on this, see [Contributing to the CSP reference articles](https://learn.microsoft.com/en-us/windows/client-management/mdm/contribute-csp-reference).

PR #11874, opened 2024-03-01
Update waas-restart.md

scottmca:
Clarifying the max active hours length.  This was based on customer support case where CSS could not figure why this policy wasn't working.  The screenshot shows 12 in the group policy help but that was only for 1607/WS2016.  After that the max is 18 hours.  This is not clearly documented anywhere

### DIFF
--- a/windows/deployment/update/waas-restart.md
+++ b/windows/deployment/update/waas-restart.md
@@ -69,6 +69,8 @@ To configure active hours using Group Policy, go to **Computer Configuration\Adm
 
 ![Use Group Policy to configure active hours.](images/waas-active-hours-policy.png)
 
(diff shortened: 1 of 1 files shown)


windows/deployment/update/waas-restart.md
```diff
@@ -69,6 +69,8 @@ To configure active hours using Group Policy, go to **Computer Configuration\Adm
 
 ![Use Group Policy to configure active hours.](images/waas-active-hours-policy.png)
 
+Note:  The max active hours length for Windows 10 1607 and Windows Server 2016 is 12.  Later versions support max active hours length of 18
```
mestew, line 72:
```suggestion
> [!Note]
>  The max active hours length for Windows 10, version 1607 and Windows Server 2016 is 12.  Later versions support max active hours length of 18 hours.
```

PR #11876, opened 2024-03-06
Update configure-use-of-passwords-for-removable-data-drives.md

Center-Deal:
Grammar correction. "the complexity [of] the password"


paolomatarazzo:
Hello @Center-Deal and thank you for catching it.

PR #11877, opened 2024-03-12
Remove external documentation links for testbase

v-shengpan:
## Why

- Closes #[Issue Number]


aczechowski (approved):
Discussed via email with Justin et al

PR #11884, opened 2024-04-08
Remove extra quote for Set-CIPolicyIdInfo PolicyId argument

forderud:
The command fails to run in PowerShell due to this extra quote.

## Why

- Closes #[Issue Number]


jgeurten (approved):
Good catch. Thanks for fixing this

PR #11886, opened 2024-04-19
Rename Windows 11 Family to Home

IsaacHatton:
## Why

I might be misunderstanding the article, but I believe someone has referred to Windows 11 Home as Windows 11 Family here, and if so as far as I know this is incorrect as that's not the name used for the product.

If I am misunderstanding this, please can someone update the article to be more clear about this, as the wording Windows 11 Family isn't used anywhere else.

## Changes

Renames Windows 11 Family to Windows 11 Home


DHB-MSFT:
@IsaacHatton  Thanks for bringing this to our attention. We've found a few other places that have incorrect references to Family, so we'll get all of these fixed.

DHB-MSFT:
@IsaacHatton  We've updated all the "Family" references to "Home" in a separate pull request and those changes are now live (as of Monday 22 April 2024, 10:45 am PDT).  Thanks again for bringing this to our attention.

PR #11888, opened 2024-04-25
Added one more item to the Audit mode policy changes

HotCakeX:
## Description

Added info about 8 Required:EV Signers which applies to Audit mode policies as well.

## Why

Because that info was missing from the document

## Changes

Added this text to the **Audit mode policies can change the behavior for some apps or cause app crashes** section

> Option **8 Required:EV Signers** is always enforced regardless of whether the policy is 
> deployed in enforced mode or in audit mode. This requirement also persists whether the
> policy utilizes the UMCI option or is exclusively applicable to kernel-mode drivers.


HotCakeX:
@jgeurten @jsuther1974 Would you please review? Thank you in advance 🙏

jsuther1974 (changes requested):
I don’t support adding this to the topic. Option 8 is *not supported*, full stop. That is already documented in https://learn.microsoft.com/en-us/windows/security/application-security/application-control/windows-defender-application-control/design/select-types-of-rules-to-create#table-1-windows-defender-application-control-policy---policy-rule-options. It doesn’t make sense to have a more specific note in the known issues for a scenario that isn’t supported in the first place. @aczechowski

HotCakeX:
Okay, sorry, i'll close it.

PR #11889, opened 2024-04-25
Removed notice about not using UTF-8 encoding

HotCakeX:
## Description

The UTF-8 encoding can totally be used for certificate subject names when deploying a signed WDAC policy and no problem such as boot failures occurs as a result of that.

I've tested this for long periods of time and continue to use it daily.

## Why

Because the notice is no longer applicable or necessary to be there. I've recorded a complete video to prove this and **you can watch it here:** 👇
https://www.youtube.com/watch?v=ae41WVQuER0

In the video I show that the certificate is using UTF-8 encoding only, for its subject name and I perform multiple restarts after deploying the signed WDAC policy to show that no boot failure occurs.

I remember this was true around a year ago and we shouldn't have used UTF-8 but since few months ago it's no longer the case.

## Changes

It removes this part

image


jgeurten (approved):

jgeurten:
Thanks @HotCakeX, it looks like this was fixed in the November 2022 update. Approving your changes

HotCakeX:
@jgeurten Thank you so much!

PR #11898, opened 2024-05-16
The Unsigned policy rule option is valid for Supplemental types

HotCakeX:
## Description

The rule option `6 Enabled:Unsigned System Integrity Policy (Default)` is valid for supplemental policies, but the document says no.

## Why

In fact, the `6 Enabled:Unsigned System Integrity Policy (Default)` **is required and must exist in supplemental policies** that are not signed, without that rule option in the supplemental policy whose base policy is also unsigned, would throw an error by the `ConvertFrom-CiPolicy` cmdlet:

```
ConvertFrom-CIPolicy: Signed policy is required, but no UpdatePolicySigner is specified.
```


jgeurten:
Good catch, @HotCakeX

HotCakeX:
Thank you 🙏

PR #11902, opened 2024-05-22
Added PowerShell command for verifying signed binaries and removed leftover text about encoding

HotCakeX:
## Description

Added a new quick PowerShell native way to verify the signed policy binary file.

The file is read as a byte array because the [SignedCms.Decode()](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.pkcs.signedcms.decode) method expects a byte array as input, it then returns an array of `X509Certificate2` objects that represent the certificate(s) used to sign the `.CIP` binary file.

Also, when reading the document, I found some leftover mentions of encoding in the warning section that no longer exists so removed them too.

## Why

* The PowerShell way produces objects instead of strings which is great for the output to be consumed by other commands.
*  It provides a lot more information than cerutil.exe
* Since all other commands in the page are in PowerShell, this offers more consistency.

### Added this

```powershell
$CIPolicyBin = 'path to signed policy file'
Add-Type -AssemblyName 'System.Security'
$SignedCryptoMsgSyntax = New-Object -TypeName System.Security.Cryptography.Pkcs.SignedCms
$SignedCryptoMsgSyntax.Decode([System.IO.File]::ReadAllBytes($CIPolicyBin))
$SignedCryptoMsgSyntax.Certificates | Format-List -Property *
```

### Removed this

```markdown
and encoding for certificate fields, like 'subject common name' and 'issuer common name'
```

### DIFF
--- a/windows/security/application-security/application-control/windows-defender-application-control/deployment/use-signed-policies-to-protect-wdac-against-tampering.md
+++ b/windows/security/application-security/application-control/windows-defender-application-control/deployment/use-signed-policies-to-protect-wdac-against-tampering.md
@@ -111,8 +111,9 @@ certutil.exe -asn <path to signed policy file>
 
 ```powershell
 $CIPolicyBin = 'path to signed policy file'
+Add-Type -AssemblyName 'System.Security'
 $SignedCryptoMsgSyntax = New-Object -TypeName System.Security.Cryptography.Pkcs.SignedCms
-$SignedCryptoMsgSyntax.Decode((Get-Content -LiteralPath $CIPolicyBin -AsByteStream -Raw))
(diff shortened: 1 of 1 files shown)


jgeurten (changes requested):
I had to first explicitly add the System.Security assembly before running line 2 of your PS command

The Decode line throws an error that Get-Content cannot accept AsByteStream as a parameter

jgeurten (approved):
Perfect. New PS commands are successfully returning the certificate used to sign the policy. Thanks for your contribution as always!!

HotCakeX:
> Perfect. New PS commands are successfully returning the certificate used to sign the policy. Thanks for your contribution as always!!

Nice, my pleasure!

PR #11906, opened 2024-06-04
Update pluton-as-tpm.md

asbaliga:
Removed certain sections from what existed previously. 

## Why

OEMs raising concerns about deploying in certain configurations which was leading to confusion amongst OEMs and users. This simplifies the messaging to all our partners and customers.


asbaliga:
@microsoft-github-policy-service agree company="Microsoft"

PR #11915, opened 2024-06-26
Update update-other-microsoft-products.md

joeloff:
Add .NET to list of products supported by Microsoft Update

## Why
.NET is part of Microsoft Update since .NET Core 3.1 and .NET 5 and should be included.

## Changes
Just adding .NET to the list of products.

### DIFF
--- a/windows/deployment/update/update-other-microsoft-products.md
+++ b/windows/deployment/update/update-other-microsoft-products.md
@@ -53,7 +53,7 @@ The following is a list of other Microsoft products that might be updated:
 - Microsoft StreamInsight
 - Mobile and IoT
 - MSRC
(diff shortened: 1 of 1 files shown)


windows/deployment/update/update-other-microsoft-products.md
```diff
@@ -53,6 +53,7 @@ The following is a list of other Microsoft products that might be updated:
 - Microsoft StreamInsight
 - Mobile and IoT
 - MSRC
+- .NET
```
jamshedd, line 56:
```suggestion
- .NET (also known as .NET Core)
```

jamshedd (approved):

PR #11919, opened 2024-07-01
Update policy-csp-search.md, removing Pro from DoNotUseWebResults

CAMSFT:
## Description

Removing Pro SKU from DoNotUseWebResults. After code review and internal email discussion with the owner of this component it was decided that the documentation should be updated to accurately reflect the SKUs where this policy works.

## Why

This policy has never worked on the Professional version of Windows. When the policy was introduced, it was restricted to specific SKUs and did not include Pro.

## Changes

Removed Pro from DoNotUseWebResults.


CAMSFT:
#assign-reviewer:vinaypamnani-msft

CAMSFT:
After further discussion, the change made in this pull request is not the right way to document this problem. Vinay has a better idea. Closing request.

PR #11921, opened 2024-07-04
Update deploy-enterprise-licenses.md

HMR18:
Device-based licensing doesn't work with Enterprise license, for example: E3 &E5, on the Windows Operating System.
- For example: upgrade From Windows 10 Pro to Windows 10/11 Enterprise using device-based licensing will not work, Enterprise _per user_ licensing is required for the Windows Operating System.

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/deployment/deploy-enterprise-licenses.md
+++ b/windows/deployment/deploy-enterprise-licenses.md
@@ -28,7 +28,8 @@ These activation features require a supported and licensed version of Windows Pr
 - Enterprise E3 in CSP.
 - Automatic, non-KMS activation also requires a device with a firmware-embedded activation key.
 - Subscription activation requires Enterprise _per user_ licensing. It doesn't work with _per device_ licensing.
-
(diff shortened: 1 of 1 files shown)


frankroj:
Hello @HMR18. Thank you for the contribution. However, I am not sure if what you added is needed. The article already has the following statement:

- Subscription activation requires Enterprise per user licensing. It doesn't work with per device licensing.

Can you clarify what your additions add to the article that isn't already covered by the above statement? Note that this article specifically is about subscription activation and not enterprise licenses in general, as mentioned in the first line of the article:

> This article describes how to deploy Windows Enterprise E3 or E5 licenses with subscription activation or Enterprise E3 in CSP and Microsoft Entra ID.

Although your statement might be correct, it doesn't apply to this article or feature since it's already stated that this feature doesn't support device based subscription.

Please let me know.

HMR18:
Hi Frank.

Thank you for your reply.
I just wanted to make it clearer.
But i guess this is covers it.

Thank you very much.

frankroj:
Thanks again @HMR18. I will go ahead and close this PR.

PR #11926, opened 2024-07-31
Update policy-csp-update.md

AbhinavMSFT:
Update Target version CSP section as productVersion is also required for TargetVision policy to work

## Why

- Closes #[Issue Number]


aczechowski:
Thanks @AbhinavMSFT for the contribution. I made a slight tweak to make sure it's in the _editable_ region of the policy. For future reference when editing CSP articles, see [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference). Since most of these pages are automated from the DDF and ADMX source files, changes outside of the editable regions will be overwritten on the next cycle.

Also, it looks like your account is already onboarded correctly so that you can use the _private_ repository. Next time, you can either use the Learn Editor (aka.ms/savetime) or switch to review.learn.microsoft.com before you select the edit option.

PR #11928, opened 2024-08-02
added an FAQ about error code 0x80d03002

narkissit:
This error code is observed when bypass is set and some content provider doesn't have a fallback.

## Why

- Closes #[Issue Number]


narkissit:
@cmknox please review

mestew:
spoke to @cmknox she approved as well.

PR #11930, opened 2024-08-09
Fix typo ("reecovery")

vivid-pixel:
I am just removing a single letter 'e' from the word 'recovery' (misspelled as 'reecovery').


aczechowski (approved):
Thanks for catching that, and thank you for your first contribution to Windows docs! 🎉

PR #11932, opened 2024-08-19
Update applications-that-can-bypass-wdac.md to fix MSBuild.exe original filename case

joeltuckwell:
## Description

Fixed the filename for FileName=MSBuild.exe to match the Original Filename in the file metadata, rather than just the file name on disk. 

## Why

This fixes case sensitive blocklist rules in security products using Original Filenames built off of this XML.

- Closes #[Issue Number]

## Changes

Changed MSBuild.Exe to MSBuild.exe to match the case for the file on disk.


jgeurten (approved):

aczechowski:
Thanks @joeltuckwell for the quick edit and thank you for your first contribution to Windows docs! 🎉

PR #11934, opened 2024-09-04
Update policy-csp-localpoliciessecurityoptions.md

cgerke:
Some policy settings can cause issues in some Windows Autopilot scenarios. This is a known issue and should be communicated to users of Autopilot.

## Why

I implemented this policy today when migrating GPO from traditional on-prem infrastructure to Intune while setting up Windows 11 with Autopilot. Intune allowed me to set this policy without any warning and the documents do not mention anything about avoiding it when using Autopilot pre-provisioning. I'd like save others from this headache. Rather than Intune Administrators reading troubleshooting documents after something is broken, we should warn them before they break it.

## Changes

Highlights the policy with an important note explaining conflicts than can occur when using the policy with Autopilot Pre-provision.


cgerke:
All done, ETA for a commit?

aczechowski:
Hi @cgerke, thanks for making the edit (and your first to Windows docs!) but we have automation for much of the CSP documentation. So in its current form, we can't accept the contribution otherwise it will be overwritten the next time the automation runs. Can you update to move your edits into the "editable" sections? For more info, see [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference).

PR #11935, opened 2024-09-06
Update windows-autopatch-prerequisites.md

JrndD:
Article just wrote that F3 was a eligible license. As there are multiple F3 licenses available, I assume that it is the Microsoft 365 F3 which grants the eligibility for autopatch.

## Why

- Closes #[Issue Number]


tiaraquan:
@JrndD  Thanks for your feedback. We're updating this article soon. Thank you!

PR #11936, opened 2024-09-10
Update waas-delivery-optimization-faq.yml

narkissit:
updating with *.windowsupdate.com becoming required (used for Defender related payload and drivers.)

## Why

- Closes #[Issue Number]


aczechowski (approved):
Thanks @narkissit !

PR #11937, opened 2024-09-14
Update usmt-recognized-environment-variables.md

eccoisle:
## Description
The following paths:
 
`C:\Program Files(86)`
`C:\Program Files(86)\Common` 

are referenced in explanations for the "CSIDL_PROGRAM_FILESX86" and "CSIDL_PROGRAM_FILES_COMMONX86" environmental variables.

I'm assuming this is an error, since these paths don't exist on a clean install of Windows 11, and placing Program Files(86) between two percent symbols: `cd %Program Files(86)%` returns `The system cannot find the path specified.`.

Additionally, the explanations for the "PROGRAMFILES(X86)" and "COMMONPROGRAMFILES(X86)" environmental variables reference the correct paths: 

`C:\Program Files (x86)`
`C:\Program Files (x86)\Common Files`

If this is intentional, I think the page would benefit from having an explanatory footnote or something similar since a cursory Google search doesn't return anything helpful.

## Why

I believe that there's an error in the documentation.

## Changes

Changed the explanations of "CSIDL_PROGRAM_FILESX86" and "CSIDL_PROGRAM_FILES_COMMONX86" to `C:\Program Files (x86)` and `C:\Program Files (x86)\Common Files`, respectively.


aczechowski (approved):
Thanks @eccoisle for catching that and submitting the edit

PR #11950, opened 2024-10-14
Update policy-csp-fileexplorer.md

StevenHosking-Msft:
## Why

Have submitted this PR to enable the ability for customers to limit folder access to only single folders.

- Closes #[Issue Number]

## Changes

expanded the table to include the digits for each of the individual folders.


aczechowski:
followed up with Steven via email

reference [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference).

PR #11952, opened 2024-10-17
Fixed typo and policy name

HotCakeX:
## Description

Fixing a typo and policy name

## Why

There was a typo "WADC". 

## Changes

It was previously correctly set to "WDAC", but at some point it was changed to "WADC" which is a typo.
Instead of setting it to "WDAC", i set it to "App Control for Business" to follow the new naming convention.


aczechowski (approved):
Thanks @HotCakeX for catching that. I checked with @vinaypamnani-msft who wanted it reduced to just App Control.

PR #11957, opened 2024-10-25
Add MCC DHCP Option

bryandam:
## Description
Added a section here, possibly incomplete by your standards, that call out that there's a DHCP option to point endpoints at the MCC server. I think that's important to call out here in the MCC docs.

## Why

It's the docs for MCC but they were incomplete in terms of the options available to point endpoints at MCC.

## Changes
See Description

### DIFF
--- a/windows/deployment/do/mcc-enterprise-appendix.md
+++ b/windows/deployment/do/mcc-enterprise-appendix.md
@@ -111,6 +111,10 @@ You can either set your Connected Cache IP address or FQDN using:
 
    :::image type="content" source="./images/ent-mcc-group-policy-hostname.png" alt-text="Screenshot of the Group Policy editor showing the Cache Server Hostname Group Policy setting." lightbox="./images/ent-mcc-group-policy-hostname.png":::
 
+#### DHCP Option 235
+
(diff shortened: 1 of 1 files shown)


bryandam:
I did smile whence I came upon the owner of this doc.

Thanks for the swift reply. Feel free to nuke this PR from orbit then if that's what makes sense. With the public preview coming in 5 days I just thought I'd make sure these docs made some reference to indicates that hardcoding a reg key isn't the only option.

So truly; reject this PR if the re-arranged docs cover both config options.

albert849 (approved):

aczechowski:
closing on Meghan's behalf, should be addressed by recent updates. @bryandam if not, please just open a new PR, it'll be easier to do a new one with the new file than trying to reopen this and deal with the merge conflict. 😄

bryandam:
Thanks @aczechowski; looks like you're just going to point at the docs for setting DOCacheHost or DOCacheHostSource which calls out the DHCP option.

Admittedly, I think it might be a _bit_ confusing in that your link is called 'Configure X policies with Intune' but you link to the DO docs instead of the Intune doc's [Delivery Optimization settings in Microsoft Intune](https://learn.microsoft.com/en-us/mem/intune/configuration/delivery-optimization-windows) page. Yet, that page doesn't cover the HostSource setting so maybe that wouldn't work for that reason.

PR #11971, opened 2024-12-04
Update journey-step-3.md

hedbergtech:
Proposing to use New-Guid instead, it will generate a "complex" password easier than having to rely on a custom PSFunction.

## Why

- Closes #[Issue Number]

### DIFF
--- a/windows/security/identity-protection/passwordless-strategy/journey-step-3.md
+++ b/windows/security/identity-protection/passwordless-strategy/journey-step-3.md
@@ -72,20 +72,8 @@ Modify the **userId** variable of the script to match your environment (first li
 ```azurepowershell-interactive
 $userId = "<UPN of the user>"
 
-function Generate-RandomPassword{
-    [CmdletBinding()]
-    param (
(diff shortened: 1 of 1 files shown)


hedbergtech:
@microsoft-github-policy-service agree [company="Truesec"]

hedbergtech:

paolomatarazzo (commented):
I don't think this would be a good idea. The generated GUID is not "complex", other that length.

hedbergtech:
On the contrary, a guid always contains a hyphen, which is in fact a "special" character, satisfying the complexity requirement. Changing the character to upper could also ve enforced.

hedbergtech:
Making it 3/4 ;). Just a suggestion, totally fine with not having the doc state it

paolomatarazzo:
fair, `-` is a special character. Still, most scenarios require upper and lower case too. Since the script has been approved by identity PMs I'm inclined to leave it as-is. Any specific concerns about the function itself? This is a sample script that can be used as a reference...

PR #11972, opened 2024-12-05
Add notice on account creation limitations.

prk0ghy:
## Why

- Closes #[Issue Number]


aczechowski (approved):
Thanks @prk0ghy for the edit, and thank you for your first for Windows content! 🎉 I made a few editorial revisions, let me know if you have any concerns with the changes.

Once you agree to the contributor license agreement, I can merge the change.

p.s. this CSP is an older style, but FYI most of the other CSP content is now automated based on the source DDF and ADMX files. So contributing to that content isn't as wide open as this one. For more information, see https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference.

PR #11973, opened 2024-12-11
Add `ScheduledInstallEveryWeek` :calendar:

blakedrumm:
## Description
Add `ScheduledInstallEveryWeek` as this parameter is not documented.

## Why
Missing this information in the public documentation. This registry key can be set via gpedit:
image

## Changes
Adds optional setting `ScheduledInstallEveryWeek` and its parameters to the documentation.


mestew:
@blakedrumm Thanks for the contribution! I'm merging this change in.

PR #11974, opened 2024-12-12
fixed Shell Launcher Powershell script on configuration/shell-launcher/shell-launcher/index.md

GiamBoscaro:
## Description

The Powershell script shown in the docs to apply the Shell Launcher configuration is not working. The configuration cannot be applied, and if error happens the script cannot retrieve the errors from the Windows Event Viewer.

## Why

Why? I am not a Windows expert at all and I have wasted days trying to setup the Shell Launcher. I have finally made it and I have figured that the documentation was actually wrong in multiple ways. I would like to update the docs to avoid this kind of trouble for other people.

## Changes

* The XML configuration must be set into the `ShellLauncher` property and not `Configuration` property of the `CimInstance`.
* `$eventLogFilterHashTable` was undefined and the script would break if some errors were to happen while applying the Shell Launcher configuration.


paolomatarazzo:
Thank you for your contribution, @GiamBoscaro. The page provided a template, while the quickstart page has a working example with the correct variable name. You are correct, the variable name in this page is not correct, therefore I approve your proposed change.

paolomatarazzo (approved):

PR #11980, opened 2024-12-27
Update windows-autopatch-faq.yml for better translation

BasileJoly:
The french translation is very confusing , i adjust to have a clear explantion in every language

## Why

- Closes #[Issue Number]


aczechowski:
Thank you @BasileJoly for the contribution and catching that issue with the translation. I can see it's also mistranslated in Spanish, but the German seems to be ok. So I'm going to file a bug to our localization team and let them take a deeper look. I'll close this pull request.

aczechowski:
for my own future searchability: bug ID 835955

PR #11982, opened 2025-01-06
Clarified that CiTool is included in Windows Server 2025 too

HotCakeX:
Clarified that CiTool is included in Windows Server 2025 too and that it's not only limited to 22H2.


aczechowski:
I think this is pretty straight forward, so just merging it. @jgeurten FYI

PR #11985, opened 2025-01-08
Updated the known issue notice in App Control policy deployment page

HotCakeX:
I updated this [once](https://github.com/MicrosoftDocs/windows-itpro-docs/pull/11954) before on a different page but just noticed it exists in here too. It clarifies that the issue has been fixed and users on 24H2 or later can deploy it normally.


aczechowski:
Another instance where this seems very straightforward based on previous edit, so I'm going to merge it. Thanks @HotCakeX! FYI @jgeurten

PR #11986, opened 2025-01-08
Update customize-boot-image.md

thehound8734:
## Why

- Closes #[Issue Number]


coreymheld (approved):

coreymheld (commented):
How do I use a duo link flash drive with the non working laptop

frankroj:
Thank you @thehound8734 for the contribution. However, the change you suggested doesn't really fix anything or change the context, so I don't think it's necessary. Can you provide some context as to why you want this change or why you think this change is necessary?

frankroj:
Hello @coreymheld. This isn't a support forum but instead where customer-initiated changes to docs can be suggested. I would post your question in a support forum where it can be properly answered. Thank you.

coreymheld:
> Thank you @thehound8734 for the contribution. However, the change you suggested doesn't really fix anything or change the context, so I don't think it's necessary. Can you provide some context as to why you want this change or why you think this change is necessary?

Cryptocurrency I just need some help trying to figure out where I sent it to or where it's at or if I have to implement some other transaction

aczechowski:
Closing

PR #11991, opened 2025-01-23
Update policy-csp-experience.md

TCGL23:
AllowedWindowsSpotlight values and decriptions do not make sense as its not enable/disable its block or allow

## Why

- Closes #[Issue Number]


aczechowski:
Thanks @TCGL23 for submitting the edit, although unfortunately we won't be able to accept it. Most of the CSP documentation is now automated from the DDF and ADMX source files. So even if we took this change now, it would reset in a few weeks when the process runs. For more information, see [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference). There's information in that article about how to file feedback via the Feedback Hub to share with the engineering team.

PR #12004, opened 2025-02-11
Update firewall-csp.md

janasebaali:
This PR fixes the description of the "enabled" field for firewall and Hyper-V rules.
The rules are enabled by default if no value is specified for the "Enabled" field. However, the descriptions state that the rules are disabled if no value is specified. This PR fixes the description to match the behavior.


aczechowski:
Thanks @janasebaali for submitting the edit, although unfortunately we won't be able to accept it. Most of the CSP documentation is now automated from the DDF and ADMX source files. So even if we took this change now, it would reset in a few weeks when the process runs. For more information, see [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference). There's information in that article about how to file feedback via the Feedback Hub to share with the engineering team.

PR #12015, opened 2025-02-25
Update windows-autopatch-windows-feature-update-policies.md

UCOwner:
## Why
In the past, we have continued updating our service-managed policies—global default and default release—for Windows 10 to the minimum supported minor version. However, with Windows 10 reaching its end of support as a major version in October, we have two options: either stop updating these policies or uplift them to Windows 11. After extensive deliberation, we believe stop uplifting those policies is less disruptive for our customers. Here are some reasons:
 
1, Automatically upgrading to a major version poses risks to customers; some devices may not be eligible for the upgrade, and thus we cannot ensure the upgrade will work for all devices. 
2, We aim to move away from default policy behaviors managed on the service side. Custom logic is currently in place to ensure they operate correctly. With the new AG group anchor policies, the use case for default policies has decreased, making it less worthwhile to maintain them.
3, We are developing new Graph APIs that includes automatic approval concepts. It would be more efficient to utilize these APIs rather than maintaining custom logic. 
The goal will still be to allow customers to stay on a supported build. We are planning to let customers leverage the automatic policies once that work is done in the future. For now, they can leverage anchor policies which will also be going out as part of no FA.  THis change is to go out when no FA GA's

- Closes #[Issue Number]


UCOwner:
@tiaraquan for when No FA goes GA

tiaraquan:
Closing PR as I will create a different one with all relevant info for this release.

PR #12023, opened 2025-03-21
Update deprecated-features-resources.md

mapalko:
Added a note about NTLMv1 removal including limitations around removal related to other capabilities using the same primitives (i.e. MSCHAPv2).

## Why

- There are some usages of NTLMv1 primitives which are still used in the OS. The current documentation is not clear about this, which is why we are adding a clarifying statement. 

## Changes

Updating NTLM deprecation resource notes.


aczechowski (approved):
Reviewed with @mestew, approving.

PR #12025, opened 2025-04-01
Update enable-virtualization-based-protection-of-code-integrity.md

justingross-msft:
This is a known issue in Azure as shown here, https://supportability.visualstudio.com/AzureIaaSVM/_wiki/wikis/AzureIaaSVM/1763930/VBS-Enabled-But-Not-Running-After-Install-HyperV_Windows. Having this note will hopefully allow customers to set this correct or have us share this page when the issue is seen.

## Why
We see these cases all the time and there is no documentation to reference for customers.

- Closes #[Issue Number]

## Changes

Just added a section about Azure VMs and the secure boot setting to use.


windows/security/hardware-security/enable-virtualization-based-protection-of-code-integrity.md
```diff
@@ -22,6 +22,7 @@ appliesto:
 >
 > - Memory integrity is sometimes referred to as *hypervisor-protected code integrity (HVCI)* or *hypervisor enforced code integrity*, and was originally released as part of *Device Guard*. Device Guard is no longer used except to locate memory integrity and VBS settings in Group Policy or the Windows registry.
 > - Memory integrity works better with Intel Kabylake and higher processors with *Mode-Based Execution Control*, and AMD Zen 2 and higher processors with *Guest Mode Execute Trap* capabilities. Older processors rely on an emulation of these features, called *Restricted User Mode*, and will have a bigger impact on performance. When nested virtualization is enabled, memory integrity works better when the VM is version >= 9.3.
+> - Azure VMs do not support memory integrity where **Secure Boot with DMA** is selected. If this is selected, VBS will show as enabled but not running. For this reason, please make sure to choose **Secure Boot** only using one of the methods below.
```
jsuther1974, line 25: Should we make this a more general recommendation? I’ve never quite understood why this was ever added as an option because it doesn’t mean you’re turning on DMA but simply limiting the policy to only apply where DMA is available.

jsuther1974 (approved):

PR #12028, opened 2025-04-09
Update policy-csp-admx-dnsclient.md

jdmu:
mistake on example of a fully qualified DNS record. Missing trailing dot '.'

## Why

- Closes #[Issue Number]


aczechowski:
Thanks @jdmu for submitting the edit, although unfortunately we won't be able to accept it. Most of the CSP documentation is now automated from the DDF and ADMX source files. In this specific example, that description comes from the ADMX, so you'll see the same text in the associated group policy. So even if we took this change now, it would reset in a few weeks when the process runs. For more information, see [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference). There's information in that article about how to file feedback via the Feedback Hub to share with the engineering team.

Another alternative is to add a note into the editable region of this setting to clarify/correct what's in the description. If you look at the comments in the markdown, you'll see the editable sections clearly defined. Feel free to reopen this PR or create a new one if you want to go that route.

PR #12030, opened 2025-04-14
Update policy-csp-deviceguard.md

sawft99:
## Description
- Changes to `./Device/Vendor/MSFT/Policy/Config/DeviceGuard/EnableVirtualizationBasedSecurity` policy

## Why
- Per Windows Event Log, the Pro edition of Windows does not support this

image

## Changes

- Changed 'EnableVirtualizationBasedSecurity' policy to show that Pro version of Windows does not support this


aczechowski:
Thanks @sawft99 for submitting the edit, although unfortunately we won't be able to accept it. Most of the CSP documentation is now automated from the DDF and ADMX source files. In this specific example, those platform applicability tables come from the DDF. So even if we took this change now, it would reset in a few weeks when the process runs. For more information, see [Contributing to the CSP reference articles](https://learn.microsoft.com/windows/client-management/mdm/contribute-csp-reference). There's information in that article about how to file feedback via the Feedback Hub to share with the engineering team.

PR #12042, opened 2025-06-02
Update windows-autopatch-edge.md

VladAutopatch:
The mmd.config definition for the policy label did not contain Microsoft. One of our big customers raised a concern between the name in documentation and the actual policy within the tenant.

## Description
There's a mismatch between the policy label in docs and in the service configuration we used to push.

## Why

Customer confusion

- Closes 2504150050003720 (DFM case number)

## Changes

Thanks for contributing to Microsoft technical content!


VladAutopatch:
[like]  Vlad Neacsu reacted to your message:

PR #12045, opened 2025-06-10
Update policy-csp-userrights.md

aidbish:
The Denylocallogon description was for services, this restores the description from a few years back.

## Why

- Closes #[Issue Number]


vinaypamnani-msft:
Thanks for the edit @aidbish - CSP docs are automated and we have an active bug to fix this description, but I'm going to accept this PR and create an override for this node so changes are persisted.

PR #12052, opened 2025-07-11
More clearly define when ESU license becomes active

4rm:
Instead of "now", indicate that an ESU license can only be activated once regular support for Win10 ends, and ESU coverage begins.

## Why

- We have users requesting ESU licenses to be applied before the coverage start date, because the KB says they can be applied "now". [Other KBs](https://learn.microsoft.com/en-us/windows/whats-new/extended-security-updates#when-will-the-esu-offer-be-available-for-licensing) accurately describe when the license can be applied. 

## Changes

- Replace "now" with more accurate indicator of when license can be applied: "once the defined ESU coverage period begins"


mestew:
@4rm thanks much for the input. We're currently editing this for clarity. You should be able to both activate and verify the key is activated now so it's in place before an update released under the ESU program actually requires it . 

Closing this PR since an edit is already in progress.

